PR #117, opened 2019-12-19
Fix configuration for public classes

### LINKED ISSUE #115: 'MA0053.public_class_should_be_sealed = false' in .editorconfig seems ignored
I added this line in my .editorconfig ```MA0053.public_class_should_be_sealed = false``` .
I have a class like this ```public class SearchResult { }``` .

Even if I added the line, the analyzer still detect that my class needs to be sealed.

Expected behavior would be public classes are no more reported.

I have both the Vsix installed and the package referenced in my project.
I'm using VS2019 16.4.1

DomLatr:
fix #115


meziantou:
Superseded by #118

PR #134, opened 2020-03-08
Add DoNotUseElseWhenIfBlockJumpsAnalyzer and its tests

louis-z:
Add an analyzer that detects an `else` block whose `if` block ends with a jump statement.

_A Fixer class could be added, either as part of this PR or an ulterior one._  @meziantou, let me know what you prefer.


src/Meziantou.Analyzer/Rules/DoNotUseElseWhenIfBlockJumpsAnalyzer.cs
```diff
@@ -0,0 +1,72 @@
...
+
+            context.RegisterSyntaxNodeAction(AnalyzeElseClause, SyntaxKind.ElseClause);
+        }
+
+        private static void AnalyzeElseClause(SyntaxNodeAnalysisContext context)
+        {
+            var elseClauseSyntax = (ElseClauseSyntax)context.Node;
+
+            var ifStatementSyntax = elseClauseSyntax.Parent;
+            var ifStatementNodes = ifStatementSyntax.ChildNodes();
```
meziantou: When working with the syntax tree, you should be very careful. Most of the time, the code you analyze is invalid (it doesn't even compile). So, be sure to check the parent exists and is actually an if statement (`IfStatementSyntax`). Then use the `Statement` property instead of `var statementOrBlock = ifStatementNodes.Skip(1).Take(1).First();` (also check it is not null)

src/Meziantou.Analyzer/Rules/DoNotUseElseWhenIfBlockJumpsAnalyzer.cs
```diff
@@ -0,0 +1,72 @@
...
+            // the 2nd one, a StatementSyntax or a BlockSyntax (i.e. the node we are interested in)
+            // the last one, the ElseClauseSyntax.
+            var statementOrBlock = ifStatementNodes.Skip(1).Take(1).First();
+            if (IsOrEndsWithJumpStatement(statementOrBlock))
+            {
+                context.ReportDiagnostic(s_rule, elseClauseSyntax);
+            }
+        }
+
+        private static bool IsOrEndsWithJumpStatement(SyntaxNode node)
```
meziantou: I'm pretty sure you can use the control flow analysis (`SemanticModel.AnalyzeControlFlow`) to detect whether the end of the block is reachable. I think this would be more robust.

Note that there is an example in `UseConfigureAwaitAnalyzer.cs`.

tests/Meziantou.Analyzer.Test/Rules/DoNotUseElseWhenIfBlockJumpsAnalyzerTests.cs
```diff
@@ -0,0 +1,136 @@
...
+
+        [Fact]
+        public async Task IfBlockJumpsAndNoElseBlockExists_NoDiagnosticIsReported()
+        {
+            var sourceCode = @"
+class TestClass
+{
+    int Test(int value)
+    {
+        while (value-- > -10)
```
meziantou: You should try to strip down the test to keep only what is necessary to test your scenario. This allows to remove some noise.

I think the test would be the same with `while (true)`. This would also allow to change the signature of the method to `void Test()`

tests/Meziantou.Analyzer.Test/Rules/DoNotUseElseWhenIfBlockJumpsAnalyzerTests.cs
```diff
@@ -0,0 +1,136 @@
...
+        {
+            if (value < 0)
+            {
+                continue;
+            }
+            System.Console.WriteLine(""Value is still positive"");
+        }
+        if (value < -100)
```
meziantou: This looks like another case to test. Why not create another unit test for this case?

tests/Meziantou.Analyzer.Test/Rules/DoNotUseElseWhenIfBlockJumpsAnalyzerTests.cs
```diff
@@ -0,0 +1,136 @@
+﻿using System.Threading.Tasks;
```
meziantou: I think you should add more tests:

````
if (true)
{
    if (true) // nested if
        return;
}
else
{
    // This is not useless
}
````

````
if (true)
{
    return;

    The last statement is a local function (I thinks this test will fail because of the implementation of IsOrEndsWithJumpStatement)
    bool Condition() => return true;
}
else
{
    // This is useless
}
````

File: tests/Meziantou.Analyzer.Test/Rules/DoNotUseElseWhenIfBlockJumpsAnalyzerTests.cs [hunk 6]
```diff
@@ -0,0 +1,136 @@
...
+        private const string IfThatThrowsElse = @"
+class TestClass
+{
+    int Test(int value)
+    {
+        if (value < 0)
+        {
+            throw new System.ArgumentNullException(nameof(value));
+        }
+    [||]else
```
meziantou: I think this is better to detect the reported text. This would ensure the diagnostic doesn't produce large squiggles in the editor.

```suggestion
    [|else|]
```

tests/Meziantou.Analyzer.Test/Rules/DoNotUseElseWhenIfBlockJumpsAnalyzerTests.cs
```diff
@@ -0,0 +1,136 @@
...
+
+        [Theory]
+        [InlineData(IfThatContinuesElse)]
+        [InlineData(IfThatReturnsElse)]
+        [InlineData(IfThatThrowsElse)]
+        public async Task IfBlockJumpsAndElseBlockExists_DiagnosticIsReported(string sourceCode)
+        {
+            await CreateProjectBuilder()
+                  .WithSourceCode(sourceCode)
+                  .ShouldReportDiagnosticWithMessage("Do not use 'else' when 'if' block jumps")
```
meziantou: I almost never test the diagnostic message. I do it only then the content is dynamic (for instance when the message contains the name of a method/parameter). But that's ok to keep it here for the moment.

src/Meziantou.Analyzer/Rules/DoNotUseElseWhenIfBlockJumpsAnalyzer.cs
louis-z: You're right. I mostly focused on the happy path.
I now added checks to make the code more robust.

src/Meziantou.Analyzer/Rules/DoNotUseElseWhenIfBlockJumpsAnalyzer.cs
louis-z: I tried using the semantic model and it initially seemed to work... but it did not for the 'yield' statement case. :(

Here's the code that I tested:

```cs
        private static void AnalyzeElseClause(SyntaxNodeAnalysisContext context)
        {
            var elseClauseSyntax = (ElseClauseSyntax)context.Node;
            if (elseClauseSyntax is null)
                return;

            if (!(elseClauseSyntax.Parent is IfStatementSyntax ifStatementSyntax))
                return;

            var statementOrBlock = ifStatementSyntax.Statement;
            if (statementOrBlock is null)
                return;

            var result = context.SemanticModel.AnalyzeControlFlow(statementOrBlock);
            if (!result.Succeeded)
                return;

            // For most jump statements, checking that ExitPoints is not empty is enough.
            // However, for 'throw' we need to also check that !EndPointIsReachable.
            if (result.ExitPoints.Length > 0 || !result.EndPointIsReachable)
            {
                context.ReportDiagnostic(s_rule, elseClauseSyntax);
            }
        }
```

tests/Meziantou.Analyzer.Test/Rules/DoNotUseElseWhenIfBlockJumpsAnalyzerTests.cs
louis-z: I created a few more tests, some of which inspired by your comments.

tests/Meziantou.Analyzer.Test/Rules/DoNotUseElseWhenIfBlockJumpsAnalyzerTests.cs
louis-z: After making this change, my tests were failing. Reverted.

src/Meziantou.Analyzer/Rules/DoNotUseElseWhenIfBlockJumpsAnalyzer.cs
meziantou: Which `yield` case? `yield return` is not an exit point; `yield break` is.

tests/Meziantou.Analyzer.Test/Rules/DoNotUseElseWhenIfBlockJumpsAnalyzerTests.cs
meziantou: `context.ReportDiagnostic(s_rule, elseClauseSyntax);` should select the whole else block. I this you should change it to specify the text span to report.

louis-z:
Thanks for the various hints, @meziantou. They were very enlightening. 👍

tests/Meziantou.Analyzer.Test/Rules/DoNotUseElseWhenIfBlockJumpsAnalyzerTests.cs
```diff
@@ -0,0 +1,224 @@
...
+    {
+        int value = -1;
+        while (true)
+        {
+            if (value < 0)
+            {
+                value++;
+                yield return value;
+            }
+        [||]else
```
meziantou: Removing this else would change the behavior of the code. `yield return` doesn't exit a method. You can test the following method to confirm it:

````c#
IEnumerable<int> A()
{
    yield return 0;
    yield return 1;
    yield return 2;
}
````

You can keep this test but remove the `[||]` and add a new test with `yield break`

Comment (louis-z) [reply tests/Meziantou.Analyzer.Test/Rules/DoNotUseElseWhenIfBlockJumpsAnalyzerTests.cs]: You're right, I was thinking of `yield break`. I made the change and added a test with `yield return` + a few more tests.

File: tests/Meziantou.Analyzer.Test/Rules/AvoidUsingRedundantElseAnalyzerTests.cs [hunk 9]
```diff
@@ -0,0 +1,285 @@
...
+class TestClass
+{
+    void Test()
+    {
+        var value = -1;
+        while (true)
+        {
+            if (value < 0)
+            {
+                {
```
louis-z, line 29: I created an extra block level here to validate that the jump statement need not be directly in the `if` block to be detected.

Actually, that's why I added the [`IsDirectAccess` method](https://github.com/meziantou/Meziantou.Analyzer/blob/2a27b5e87d0522d213bf92d3f5e3649ac29a013b/src/Meziantou.Analyzer/Rules/AvoidUsingRedundantElseAnalyzer.cs#L63)

meziantou: `IsDirectAccess` is only needed because of the local function. If you remove it you can see that `IsEndPointReachable` is false. This seems like a weird behavior of Roslyn. 

`IsDirectAccess` is fine with me.

src/Meziantou.Analyzer/Rules/AvoidUsingRedundantElseAnalyzer.cs
```diff
@@ -0,0 +1,77 @@
...
+            if (statementOrBlock is null)
+                return;
+
+            var result = context.SemanticModel.AnalyzeControlFlow(statementOrBlock);
+            if (!result.Succeeded)
+                return;
+
+            if (!result.EndPointIsReachable || result.ExitPoints.Any(ep => IsDirectAccess(ifStatementSyntax, ep)))
+            {
+                context.ReportDiagnostic(s_rule, elseClauseSyntax);
```
meziantou:
```suggestion
                context.ReportDiagnostic(s_rule, elseClauseSyntax.ElseKeyword);
```

This way it will report only on the else keyword instead of the whole else block. Note that you'll need to adjust the test to reflect that change (`[|else|]` instead of `[||]else`)

louis-z: I made the change to the code, but the tests still succeeded without any adjustment. 😕

meziantou: They do, but I prefer you change them to validate that the analyzer only reports the else keyword instead of the else block

tests/Meziantou.Analyzer.Test/Rules/DoNotUseElseWhenIfBlockJumpsAnalyzerTests.cs
louis-z: Same comment as below. Resolving this one.

src/Meziantou.Analyzer/Rules/AvoidUsingRedundantElseAnalyzer.cs
louis-z: Done

PR #139, opened 2020-03-16
Improve "Avoid using redundant else fixer"

louis-z:
- Improve `AvoidUsingRedundantElseFixer`
- Refactor tests a bit


src/Meziantou.Analyzer/Rules/AvoidUsingRedundantElseFixer.cs
```diff
@@ -38,75 +38,46 @@ public override async Task RegisterCodeFixesAsync(CodeFixContext context)
...
+            var ifStatementParent = ifStatementSyntax.Parent;
+            if (ifStatementParent is null)
+                return document;
 
-            // Remove else
-            root = root.ReplaceNode(ifStatement, ifStatement.WithElse(@else: null));
-            ifStatement = (IfStatementSyntax)root.FindNode(ifStatement.IfKeyword.Span);
+            var modifiedIfStatement = ifStatementSyntax
+                .WithElse(null)
+                .WithTrailingTrivia(ifStatementSyntax.GetTrailingTrivia().Add(LineFeed));
```
meziantou: Could you add a test with an empty else statement

````
void A()
{
    if (true)
    {
         return;
    }
    else
    {
    }
}
````

Should fix with the following code (no empty line after the if)

````
void A()
{
    if (true)
    {
         return;
    }
}
````

louis-z: Good idea! It turns out the fixer was still outputting the empty line after the `if` block. 🙄 
Fixing it...

Issue #80, opened 2019-08-11
New rule: Exceptions should not be thrown in finally blocks


meziantou:

louis-z:
I _(quickly)_ implemented a first version of the `DoNotThrowExceptionFromFinallyBlockAnalyzer`. Is it supposed to detect only explicit `throw` statements or must it also determine whether any call made from a `finally` block is likely to throw?

meziantou:
We cannot do advanced checks. So, I think it should only detect direct throw statements. Maybe we can skip catched exceptions.

````
try
{
}
finally
{
    throw new Exception(); // report

    try
    {
         throw new Exception(); // ok
    }
    catch
    {
    }

    try
    {
         throw new Exception(); // report?
    }
    catch (SpecificException) // doesn't catch Exception
    {
    }
}
````

louis-z:
And of course, I'll need to be smarter than just rely on `SemanticModel.AnalyzeControlFlow` and check if `EndPointIsReachable` is false... 😉

meziantou:
To be honest, I think we you just detect any throw statement in a finally block. They all seems wrong to me.
So, an implementation based on IOperation should be fine.

PR #141, opened 2020-03-17
Implement DoNotThrowFromFinallyBlockAnalyzer

louis-z:
This resolves issue https://github.com/meziantou/Meziantou.Analyzer/issues/80


src/Meziantou.Analyzer/Rules/DoNotThrowFromFinallyBlockAnalyzer.cs
```diff
@@ -0,0 +1,53 @@
...
+
+            foreach (var throwStatement in finallyBlock.DescendantNodes().Where(IsThrowStatement))
+            {
+                context.ReportDiagnostic(s_rule, throwStatement);
+            }
+        }
+
+        private static bool IsThrowStatement(SyntaxNode node)
+        {
+            return node.GetType() == typeof(ThrowStatementSyntax);
```
meziantou: You should use `node.Kind` instead of comparing types, this is faster or not dependent on the implementation.

louis-z: Good idea.

meziantou:
LGTM!

PR #142, opened 2020-03-23
Small improvements to code recently added

louis-z:
Plus:
- Delete global.json, so that developers can use any (recent) SDK version
- Bump up to 3.1.200 the SDK version used in CI

### DIFF
--- a/Global.json
+++ b/Global.json
@@ -1,6 +1,0 @@
-{
-  "sdk": {
-    "version": "3.1.100",
-    "rollForward": "latestFeature"
-  }
-}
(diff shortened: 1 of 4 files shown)


Global.json
```diff
@@ -1,5 +1,6 @@
 {
```
meziantou: The version of the sdk is fixed in the CI. So, this file is only for local environment. Instead of updating it regularly I think we can just remove it. As long as you have a recent sdk it should work.

BTW can you update the SDK in `azure-pipeline.json` to 3.1.200?

Issue #89, opened 2019-08-11
New rule: Remove useless binary comparisons with bool


meziantou:
````
bool condition;
if (condition == true) // should be if (condition)
if (condition == false) // should be if (!condition)
````

PR #143, opened 2020-03-25
Add AvoidComparisonWithBoolConstantAnalyzer, Fixer & Tests

louis-z:
Resolves https://github.com/meziantou/Meziantou.Analyzer/issues/89


src/Meziantou.Analyzer/Rules/AvoidComparisonWithBoolLiteralAnalyzer.cs
```diff
@@ -0,0 +1,51 @@
...
+
+            if (binaryExpressionSyntax.Left is null || binaryExpressionSyntax.Right is null)
+                return;
+
+            if (binaryExpressionSyntax.Left.IsKind(SyntaxKind.TrueLiteralExpression) ||
+                binaryExpressionSyntax.Left.IsKind(SyntaxKind.FalseLiteralExpression) ||
+                binaryExpressionSyntax.Right.IsKind(SyntaxKind.TrueLiteralExpression) ||
+                binaryExpressionSyntax.Right.IsKind(SyntaxKind.FalseLiteralExpression))
+            {
+                context.ReportDiagnostic(s_rule, binaryExpressionSyntax.OperatorToken);
```
meziantou: This is not true for `bool?`.

````c#
bool? a = true;
if(a == true) // compile
if(a) // does not compile

````

File: src/Meziantou.Analyzer/Rules/AvoidComparisonWithBoolLiteralAnalyzer.cs [hunk 2]
```diff
@@ -0,0 +1,51 @@
...
+            ImmutableArray.Create(SyntaxKind.EqualsExpression, SyntaxKind.NotEqualsExpression);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(s_rule);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+
+            context.RegisterSyntaxNodeAction(AnalyzeComparisonWithBool, s_comparisonExpressionKinds);
```
meziantou: For this analyzer, I would go with `IBinaryOperation`. This way you can get the type of the operands and be sure one operand if of type `bool` and the other one is a constant (true or false).

If you are not aware of IOperation, you can check my blog post: <https://www.meziantou.net/writing-a-language-agnostic-roslyn-analyzer-using-ioperation.htm>. You can also check other analyzers that use operations.

src/Meziantou.Analyzer/Rules/AvoidComparisonWithBoolLiteralAnalyzer.cs
louis-z: You're absolutely right. 😳 
I'll have another go at it.

louis-z: Made a test out of your code snippet.

src/Meziantou.Analyzer/Rules/AvoidComparisonWithBoolLiteralAnalyzer.cs
louis-z: Made changes to the Analyzer (but not the Fixer). Let me know what you think.

src/Meziantou.Analyzer/Rules/AvoidComparisonWithBoolLiteralAnalyzer.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+
+            context.RegisterOperationAction(AnalyzeBinaryOperation, OperationKind.Binary);
+        }
+
+        private static void AnalyzeBinaryOperation(OperationAnalysisContext context)
```
meziantou: That much more robust than the first implementation!

I think you can rewrite `binaryOperation.LeftOperand.Kind == OperationKind.Literal` with `ConstantValue`. I think this can also detect constant usage (if it does, you can add a new unit test :)):

````c#
const bool TrueConstant = true;

bool a = false;
_ = a == TrueConstant; // I think it would detect this
````

Here's what I would have written (not tested):

````c#
// Must have 2 operands
if (binaryOperation.LeftOperand is null || binaryOperation.RightOperand is null)
    return;

// Operands must be in the code
if (binaryOperation.LeftOperand.IsImplicit || binaryOperation.RightOperand.IsImplicit)
    return;

// Operands must be boolean
if(!binaryOperation.LeftOperand.Type.IsBoolean() || !binaryOperation.RightOperand.Type.IsBoolean())
    return;

// Report diagnostic if one operand is a constant
if (IsConstantBooleanValue(binaryOperation.LeftOperand) || IsConstantBooleanValue(binaryOperation.RightOperand))
{
    ReportDiagnostic(...);
}

static bool IsConstantBooleanValue(IOperation operation)
{
    return operation.ConstantValue.HasValue && operation.ConstantValue.Value is bool;
}
````

File: tests/Meziantou.Analyzer.Test/Rules/AvoidComparisonWithBoolLiteralAnalyzerTests.cs [hunk 4]
```diff
@@ -0,0 +1,102 @@
+﻿using System.Threading.Tasks;
+using Meziantou.Analyzer.Rules;
+using TestHelper;
+using Xunit;
+
+namespace Meziantou.Analyzer.Test.Rules
+{
+    public sealed class AvoidComparisonWithBoolLiteralAnalyzerTests
```
meziantou: You should add more tests for valid cases

````
bool a = true;
if (a) // no diagnostic

if (!a) // no diagnostic
````

File: src/Meziantou.Analyzer/Rules/AvoidComparisonWithBoolLiteralAnalyzer.cs (hunk 3 above)
Comment (louis-z) [reply src/Meziantou.Analyzer/Rules/AvoidComparisonWithBoolLiteralAnalyzer.cs]: Made changes based on your suggestions (but not quite a copy) and renamed files and classes:

BoolLiteral -> BoolConstant

File: tests/Meziantou.Analyzer.Test/Rules/AvoidComparisonWithBoolLiteralAnalyzerTests.cs (hunk 4 above)
Comment (louis-z) [reply tests/Meziantou.Analyzer.Test/Rules/AvoidComparisonWithBoolLiteralAnalyzerTests.cs]: Added a bit more test cases

File: src/Meziantou.Analyzer/Rules/AvoidComparisonWithBoolLiteralAnalyzer.cs (hunk 2 above)
Comment (louis-z) [reply src/Meziantou.Analyzer/Rules/AvoidComparisonWithBoolLiteralAnalyzer.cs]: In the end, I had to also modify the fixer.

Comment (meziantou) [reply src/Meziantou.Analyzer/Rules/AvoidComparisonWithBoolLiteralAnalyzer.cs]: That's good :)

**Comment by meziantou** [comment]
For the fixer, I think the easiest way is to send data from the analyzer to the fixer. You can send data using a dictionary in ReportDiagnostic (there is at least one example in the code base). For instance, you can set which operand to remove (left or right) and if it needs to add the unary not operator.

**Comment by louis-z** [comment]
I did as suggested in https://github.com/meziantou/Meziantou.Analyzer/pull/143#issuecomment-604704565

File: tests/Meziantou.Analyzer.Test/Rules/AvoidComparisonWithBoolConstantAnalyzerTests.cs [hunk 5]
```diff
@@ -0,0 +1,188 @@
...
+        var success = {boolLiteral} [|{equalityOperator}|] (GetSomeNumber() == 15);
+        int GetSomeNumber() => 12;
+    }}
+}}";
+            var modifiedCode = $@"
+class TestClass
+{{
+    void Test()
+    {{
+        var success = {expectedPrefix}(GetSomeNumber() == 15);
```
meziantou: In the case `var success = true == (GetSomeNumber() == 15);`, the fix should remove the parentheses: `var success = GetSomeNumber() == 15;`.

I think you need to add the `Simplifier` annotation

meziantou:
Could you rebase on master and call GenerateDocumentation.cmd?

docs/Rules/MA0073.md
```diff
@@ -0,0 +1 @@
+# MA0073 - Avoid comparison with bool constant
```
meziantou: It could be nice to add an example of a non-compliant code :)

Issue #87, opened 2019-08-11
New rule: Classes that provide "Equals(<T>)" should implement "IEquatable<T>"


meziantou:

louis-z:
@meziantou, if you want to I'd be interested in working on this issue.

meziantou:
Thanks for the PR!

PR #144, opened 2020-04-01
Implement ClassWithEqualsTShouldImplementIEquatableTAnalyzer

louis-z:
Resolves https://github.com/meziantou/Meziantou.Analyzer/issues/87

### DIFF
--- a/src/Meziantou.Analyzer/Rules/ClassWithEqualsTShouldImplementIEquatableTAnalyzer.cs
+++ b/src/Meziantou.Analyzer/Rules/ClassWithEqualsTShouldImplementIEquatableTAnalyzer.cs
@@ -50,12 +50,12 @@ namespace Meziantou.Analyzer.Rules
                 return;
             }
 
-            var iequatableType = context.Compilation.GetTypeByMetadataName("System.IEquatable`1");
-            if (iequatableType == null)
+            var genericInterfaceSymbol = context.Compilation.GetTypeByMetadataName("System.IEquatable`1");
+            if (genericInterfaceSymbol == null)
                 return;
 
-            iequatableType = iequatableType.Construct(equalsMethod.ContainingType);
-            if (equalsMethod.ContainingType.Implements(iequatableType))
+            var concreteInterfaceSymbol = genericInterfaceSymbol.Construct(equalsMethod.ContainingType);
+            if (equalsMethod.ContainingType.Implements(concreteInterfaceSymbol))
                 return;
 
             context.ReportDiagnostic(s_rule, equalsMethod.ContainingType);
--- a/src/Meziantou.Analyzer/Rules/ClassWithEqualsTShouldImplementIEquatableTAnalyzer.cs
+++ b/src/Meziantou.Analyzer/Rules/ClassWithEqualsTShouldImplementIEquatableTAnalyzer.cs
@@ -1,0 +1,51 @@
+﻿using System;
+using System.Collections.Immutable;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Meziantou.Analyzer.Rules
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public sealed class ClassWithEqualsTShouldImplementIEquatableTAnalyzer : DiagnosticAnalyzer
+    {
+        private static readonly DiagnosticDescriptor s_rule = new DiagnosticDescriptor(
+            RuleIdentifiers.ClassWithEqualsTShouldImplementIEquatableT,
+            title: "A class that provides Equals(T) should implement IEquatable<T>",
+            messageFormat: "A class that provides Equals(T) should implement IEquatable<T>",
+            RuleCategories.Design,
+            DiagnosticSeverity.Warning,
+            isEnabledByDefault: true,
+            description: "",
+            helpLinkUri: RuleIdentifiers.GetHelpUri(RuleIdentifiers.ClassWithEqualsTShouldImplementIEquatableT));
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(s_rule);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+
+            context.RegisterSymbolAction(AnalyzeMethodSymbol, SymbolKind.Method);
+        }
+
+        private static void AnalyzeMethodSymbol(SymbolAnalysisContext context)
+        {
+            var symbol = (IMethodSymbol)context.Symbol;
+
+            if (!string.Equals(symbol.Name, "Equals", StringComparison.Ordinal))
+                return;
+
+            if (symbol.IsStatic || symbol.MethodKind != MethodKind.Ordinary || symbol.DeclaredAccessibility != Accessibility.Public)
+                return;
+
+            if (symbol.Parameters.Length != 1 || !symbol.Parameters[0].Type.IsEqualTo(symbol.ContainingType) || !symbol.ReturnType.IsBoolean())
+                return;
+
+            if (symbol.ContainingType.Interfaces.Any(i => i.Name.Equals("IEquatable", StringComparison.Ordinal)))
+                return;
+
+            context.ReportDiagnostic(s_rule, symbol.ContainingType);
+        }
+    }
+}
--- a/src/Meziantou.Analyzer/Rules/ClassWithEqualsTShouldImplementIEquatableTAnalyzer.cs
+++ b/src/Meziantou.Analyzer/Rules/ClassWithEqualsTShouldImplementIEquatableTAnalyzer.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Collections.Immutable;
-using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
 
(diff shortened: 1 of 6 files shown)


louis-z:
For now, only Analyzer & Tests were provided.

With a bit more time, I'll write the Fixer too. ;)

src/Meziantou.Analyzer/Rules/ClassWithEqualsTShouldImplementIEquatableTAnalyzer.cs
```diff
@@ -0,0 +1,51 @@
...
+            if (!string.Equals(symbol.Name, "Equals", StringComparison.Ordinal))
+                return;
+
+            if (symbol.IsStatic || symbol.MethodKind != MethodKind.Ordinary || symbol.DeclaredAccessibility != Accessibility.Public)
+                return;
+
+            if (symbol.Parameters.Length != 1 || !symbol.Parameters[0].Type.IsEqualTo(symbol.ContainingType) || !symbol.ReturnType.IsBoolean())
+                return;
+
+            if (symbol.ContainingType.Interfaces.Any(i => i.Name.Equals("IEquatable", StringComparison.Ordinal)))
```
meziantou: Comparing by name is not enough. For instance, you'll match `MyNamespace.IEquatable`. Instead you should get the symbol using `context.Compilation.GetTypeByMetatadaName("System.IEquatable`1")`.
Also, you need to check the generic parameter match the current class name.

The pseudo-code should look like:

````c#
var equatableSymbol = context.Compilation.GetTypeByMetatadaName("System.IEquatable`1"); // There are lots of example in the project
if (equatableSymbol == null) // The type may not exist in the current compilation context
    return;

var equatableCurrentTypeSymbol = equatableSymbol.Construct(symbol.ContainingType); // Replace T in IEquatable<T> by the actual type
if (symbol.ContainingType.Implements(equatableCurrentTypeSymbol)) // Check the current type implements the Interface
    return;

````

File: tests/Meziantou.Analyzer.Test/Rules/ClassWithEqualsTShouldImplementIEquatableTAnalyzerTests.cs [hunk 2]
```diff
@@ -0,0 +1,58 @@
...
+    public sealed class ClassWithEqualsTShouldImplementIEquatableTAnalyzerTests
+    {
+        private static ProjectBuilder CreateProjectBuilder()
+        {
+            return new ProjectBuilder()
+                .WithAnalyzer<ClassWithEqualsTShouldImplementIEquatableTAnalyzer>();
+        }
+
+        [Theory]
+        [InlineData("public bool Equals(Test other) { return false; }", true)]
```
meziantou: I often use `throw null` to simplify tests and indicates you don't care about the method implementation. It works whatever the method type and you can use `=>`

````
public void Equals(Test other) => throw null;
public bool Equals(Test other) => throw null;
````

**Comment by meziantou** [comment]
I don't know if we want to report the diagnostic in an interface

````c#
public interface A
{
    bool Equals(A other); // Should we suggest to implement the interface?
}
````

File: src/Meziantou.Analyzer/Rules/ClassWithEqualsTShouldImplementIEquatableTAnalyzer.cs (hunk 1 above)
Comment (louis-z) [reply src/Meziantou.Analyzer/Rules/ClassWithEqualsTShouldImplementIEquatableTAnalyzer.cs]: You're right. I added your logic.

**Comment by louis-z** [comment]
Regarding https://github.com/meziantou/Meziantou.Analyzer/pull/144#issuecomment-606986767, I propose we don't generate a diagnostic for an interface. After all, there must be a reason why such an interface exists and, if so, it may have nothing to do with IEquatable's purpose (even if it does provide an Equals<T> method).

File: tests/Meziantou.Analyzer.Test/Rules/ClassWithEqualsTShouldImplementIEquatableTAnalyzerTests.cs [hunk 3]
```diff
@@ -0,0 +1,107 @@
+﻿using System.Threading.Tasks;
```
meziantou, line 1: You only test classes and interfaces. You should also test structs.

louis-z: Done

louis-z:
Do you have any hints on how to code the fixer? Can it also use the ISymbol API or is that a read-only API?

meziantou:
The ISymbol API cannot be used to generate code. You can use:

- the editor (prefered method)

    ````
    var editor = await DocumentEditor.CreateAsync(document, cancellationToken).ConfigureAwait(false);
    editor.AddBaseType(ClassDeclarationSyntax, Type);
    ````

- the syntax tree:

    ````
    ClassDeclarationSyntax.AddBaseListTypes(...); // be sure to support class and struct
    ````

louis-z:
The pipeline failure seems unrelated to this PR.

meziantou:
I'll check for the failed tests

meziantou:
The latest commit on master is green. Rebasing your branch on master should fix the build.

tests/Meziantou.Analyzer.Test/Rules/ClassWithEqualsTShouldImplementIEquatableTAnalyzerTests.cs
```diff
@@ -0,0 +1,216 @@
...
```
meziantou, line 1: It would be nice to introduce a new test with nullable reference types. There is no example in this repository but I think you can use something like:

````csharp
            var originalCode = @"
#nullable enable
class [|Test|]
{
    public bool Equals(Test? other) => throw null;
}";
            var modifiedCode = @"
class Test : System.IEquatable<Test> // I don't know what's best here IEquatable<Test> or IEquatable<Test?> (I think you can check what VS generates for equality members)
{
    public bool Equals(Test? other) => throw null;
}";
````

File: src/Meziantou.Analyzer/Rules/ClassWithEqualsTShouldImplementIEquatableTFixer.cs [hunk 5]
```diff
@@ -0,0 +1,68 @@
...
+            var title = "Implement System.IEquatable";
+            var codeAction = CodeAction.Create(
+                title,
+                ct => ImplementIEquatable(context.Document, nodeToFix, ct),
+                equivalenceKey: title);
+
+            context.RegisterCodeFix(codeAction, context.Diagnostics);
+        }
+
+        private static async Task<Document> ImplementIEquatable(Document document, SyntaxNode nodeToFix, CancellationToken cancellationToken)
```
meziantou, line 37: I think this is more robust by using the generator. Something like the following (missing some validation and null checks)

````
        private static async Task<Document> ImplementIEquatable(Document document, SyntaxNode nodeToFix, CancellationToken cancellationToken)
        {
            var semanticModel = await document.GetSemanticModelAsync(cancellationToken).ConfigureAwait(false);
            var editor = await DocumentEditor.CreateAsync(document, cancellationToken).ConfigureAwait(false);
            var generator = editor.Generator;

            var declaredSymbol = semanticModel.GetDeclaredSymbol(nodeToFix) as ITypeSymbol;

            var symbol = semanticModel.Compilation.GetTypeByMetadataName("System.IEquatable`1");
            var symbolOfT = symbol.Construct(declaredSymbol);

            var newBaseType = generator.TypeExpression(symbolOfT);

            editor.AddBaseType(nodeToFix, newBaseType);
            return editor.GetChangedDocument();
        }
````

Comment (louis-z) [reply src/Meziantou.Analyzer/Rules/ClassWithEqualsTShouldImplementIEquatableTFixer.cs:37]: Great idea! Thanks for the suggestion.

Note, however, that I used `editor.AddInterfaceType` instead of `editor.AddBaseType`.

Comment (meziantou) [reply src/Meziantou.Analyzer/Rules/ClassWithEqualsTShouldImplementIEquatableTFixer.cs:37]: That's even better :)

File: src/Meziantou.Analyzer/Rules/ClassWithEqualsTShouldImplementIEquatableTFixer.cs [hunk 6]
```diff
@@ -0,0 +1,64 @@
...
+            var declaredTypeSymbol = semanticModel.GetDeclaredSymbol(nodeToFix) as ITypeSymbol;
+            if (declaredTypeSymbol is null)
+                return document;
+
+            var genericInterfaceSymbol = semanticModel.Compilation.GetTypeByMetadataName("System.IEquatable`1");
+            if (genericInterfaceSymbol is null)
+                return document;
+
+            var concreteInterfaceSymbol = genericInterfaceSymbol.Construct(declaredTypeSymbol);
+            if (concreteInterfaceSymbol is null)
```
meziantou: This check is useless as construct cannot return null.

src/Meziantou.Analyzer/Rules/ClassWithEqualsTShouldImplementIEquatableTFixer.cs
```diff
@@ -0,0 +1,64 @@
...
+                return document;
+
+            var editor = await DocumentEditor.CreateAsync(document, cancellationToken).ConfigureAwait(false);
+            var generator = editor.Generator;
+
+            var concreteInterfaceTypeNode = generator.TypeExpression(concreteInterfaceSymbol);
+
+            editor.AddInterfaceType(nodeToFix, concreteInterfaceTypeNode.WithoutTrailingTrivia());
```
meziantou, line 64: Why do you need `.WithoutTrailingTrivia()`

src/Meziantou.Analyzer/Rules/ClassWithEqualsTShouldImplementIEquatableTFixer.cs
louis-z: Ok, I'll remove it.

src/Meziantou.Analyzer/Rules/ClassWithEqualsTShouldImplementIEquatableTFixer.cs
louis-z: Because otherwise one of the tests fails (see [line 47](https://github.com/meziantou/Meziantou.Analyzer/blob/30f6393ed294fd1a139f62e45b95e07e006b42f0/tests/Meziantou.Analyzer.Test/Rules/ClassWithEqualsTShouldImplementIEquatableTAnalyzerTests.cs#L47)), with a comment that winds up on the next line.

src/Meziantou.Analyzer/Rules/ClassWithEqualsTShouldImplementIEquatableTFixer.cs
```diff
@@ -47,10 +48,20 @@ private static async Task<Document> ImplementIEquatable(Document document, Synta
             if (genericInterfaceSymbol is null)
                 return document;
 
-            var concreteInterfaceSymbol = genericInterfaceSymbol.Construct(declaredTypeSymbol);
-            if (concreteInterfaceSymbol is null)
+            // Retrieve Nullable Annotation from the Equals method and use it to construct the concrete interface
```
meziantou, line 48: - I'm not sure we need to add the nullable annotation because `IEquatable<T>.Equals` [has the `[AllowNull]` attribute](https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/IEquatable.cs#L9). The dotnet/runtime repository seems to contain more examples without the nullable annotation than with the nullable annotation. But maybe this is because it is mostly old code. At the same time I think this is clearer to have the annotation. => I really don't have a strong opinion on it, so I think we can go with the nullable annotation as you have already implemented it.

- Instead of doing the logic in the analyzer and the fixer to find the Equals method, you should do it once. You can use an internal static method in the analyzer or store data using the properties dictionary.

louis-z: Done

Issue #146, opened 2020-04-05
MA0071 fixer fails for nested if statements containing several redundant else statements


louis-z:
The following test fails. The analyzer correctly detects 3 redundant `else` statements, but the fixer only corrects 2. Perhaps the analyzer needs to be relaunched after each modification to the syntax tree?

```cs
        [Fact]
        public async Task Test_NestedIfBlocksWithJumps()
        {
            var originalCode = @"
class TestClass
{
    void Test()
    {
        var value = -1;
        while (true)
        {
            if (value > 0)
            {
                return;
            }
            [|else|] if (value < -10)
            {
                continue;
            }
            [|else|]
            {
                if (value < 0)
                {
                    break;
                }
                [|else|]
                {
                    value++;
                }
            }
        }
    }
}";
            var modifiedCode = @"
class TestClass
{
    void Test()
    {
        var value = -1;
        while (true)
        {
            if (value > 0)
            {
                return;
            }

            if (value < -10)
            {
                continue;
            }

            if (value < 0)
            {
                break;
            }

            value++;
        }
    }
}";
            await CreateProjectBuilder()
                  .WithSourceCode(originalCode)
                  .ShouldFixCodeWith(modifiedCode)
                  .ValidateAsync();
        }
```

meziantou:
I think it has something to do with the method `FixAllProvider GetFixAllProvider()`. Maybe you need to implement a custom FixAllProvider (I've never done that before).

PR #149, opened 2020-04-09
Add a test that reproduces issue 146 and make changes to address it

louis-z:
This addresses https://github.com/meziantou/Meziantou.Analyzer/issues/146

- Add Test_SeveralNestedIfElseBlocksWithIfsThatJump_AllProblematicElsesRemoved
- Copy over
    - DocumentBasedFixAllProvider.cs, and
    - FixAllContextHelper.cs
  from https://github.com/DotNetAnalyzers/StyleCopAnalyzers
- Implement AvoidUsingRedundantElseFixAllProvider, derived from DocumentBasedFixAllProvider
- Modify AvoidUsingRedundantElseFixer so that:
    - GetFixAllProvider() returns an AvoidUsingRedundantElseFixAllProvider
    - it processes diagnostics per document, in reverse line order so as not to invalidate any of the remaining nodes to fix

### DIFF
--- a/src/Meziantou.Analyzer/Rules/AvoidUsingRedundantElseFixer.cs
+++ b/src/Meziantou.Analyzer/Rules/AvoidUsingRedundantElseFixer.cs
@@ -19,22 +19,39 @@ namespace Meziantou.Analyzer.Rules
     {
         public override ImmutableArray<string> FixableDiagnosticIds => ImmutableArray.Create(RuleIdentifiers.AvoidUsingRedundantElse);
 
-        public override FixAllProvider GetFixAllProvider() => WellKnownFixAllProviders.BatchFixer;
-
-        public override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        /// <inheritdoc/>
+        public override FixAllProvider GetFixAllProvider()
         {
-            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
-            var nodeToFix = root.FindNode(context.Span, getInnermostNodeForTie: true);
-            if (nodeToFix == null)
-                return;
+            // We can't use WellKnownFixAllProviders.BatchFixer here, as we need to fix diagnostics per document in reverse order
+            return AvoidUsingRedundantElseFixAllProvider.Instance;
+        }
 
+        public override Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
             var title = "Remove redundant else";
             var codeAction = CodeAction.Create(
                 title,
-                ct => RemoveRedundantElse(context.Document, nodeToFix, ct),
+                ct => RemoveRedundantElseClausesInDocument(context.Document, context.Diagnostics, ct),
                 equivalenceKey: title);
 
             context.RegisterCodeFix(codeAction, context.Diagnostics);
+
+            return Task.CompletedTask;
+        }
+
+        internal static async Task<Document> RemoveRedundantElseClausesInDocument(Document document, ImmutableArray<Diagnostic> diagnostics, CancellationToken cancellationToken)
+        {
+            foreach (var diagnostic in diagnostics.Reverse())
+            {
+                var root = await document.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
+                var nodeToFix = root.FindNode(diagnostic.Location.SourceSpan, getInnermostNodeForTie: true);
+                if (nodeToFix == null)
+                    continue;
+
+                document = await RemoveRedundantElse(document, nodeToFix, cancellationToken).ConfigureAwait(false);
+            }
+
+            return document;
         }
 
         private static async Task<Document> RemoveRedundantElse(Document document, SyntaxNode nodeToFix, CancellationToken cancellationToken)
--- a/src/Meziantou.Analyzer/Rules/AvoidUsingRedundantElseFixer.cs
+++ b/src/Meziantou.Analyzer/Rules/AvoidUsingRedundantElseFixer.cs
@@ -41,7 +41,7 @@ namespace Meziantou.Analyzer.Rules
 
         internal static async Task<Document> RemoveRedundantElseClausesInDocument(Document document, ImmutableArray<Diagnostic> diagnostics, CancellationToken cancellationToken)
         {
-            foreach (var diagnostic in diagnostics.Reverse())
+            foreach (var diagnostic in diagnostics.OrderByDescending(d => d.Location.SourceSpan))
             {
                 var root = await document.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
                 var nodeToFix = root.FindNode(diagnostic.Location.SourceSpan, getInnermostNodeForTie: true);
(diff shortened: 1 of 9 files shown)


src/Meziantou.Analyzer/Rules/AvoidUsingRedundantElseAnalyzer.cs
```diff
@@ -27,30 +27,43 @@ public override void Initialize(AnalysisContext context)
...
+            var elseClauses = rootNode.DescendantNodes().Where(n => n.IsKind(SyntaxKind.ElseClause)).Cast<ElseClauseSyntax>();
+            var locations = elseClauses
+                .Select(clause => AnalyzeElseClause(context, clause))
+                .Where(location => location != null);
+
+            if (!locations.Any())
                 return;
 
+            var diagnostic = Diagnostic.Create(s_rule, locations.First(), locations.Skip(1));
+            context.ReportDiagnostic(diagnostic);
```
meziantou: I don't like the fact that we report only one diagnostic for the whole file because this means a user cannot fix only one of them. Instead, they must fix all of them at once.
So, maybe this is not the right way to solve the issue. However, I think you can create a FixAllProvider that apply the same approach as here. This means:
- Report all diagnostic as before
- Implement a fix all fixer ([documentation](https://github.com/dotnet/roslyn/blob/master/docs/analyzers/FixAllProvider.md)) that apply the fix in the reverse order (you can sort them by location). 

There are a few examples of Batch Fixers in the Roslyn repository ([example](https://github.com/dotnet/roslyn/blob/master/src/Workspaces/Core/Portable/CodeFixes/FixAllOccurrences/BatchFixAllProvider.cs), [example](https://github.com/dotnet/roslyn/blob/master/src/Features/Core/Portable/CodeFixes/Suppression/AbstractSuppressionCodeFixProvider.PragmaWarningBatchFixAllProvider.cs), [example](https://github.com/dotnet/roslyn-analyzers/blob/master/src/PublicApiAnalyzers/Core/CodeFixes/DeclarePublicApiFix.cs)). This doesn't think that hard to create. However, we don't to update the code to bw able to trigger the batch fixer when needed.


Another solution is to not support fix all by returning null.

````
public override FixAllProvider GetFixAllProvider() => null;
````

Comment (louis-z) [reply src/Meziantou.Analyzer/Rules/AvoidUsingRedundantElseAnalyzer.cs]: I initially considered (albeit briefly) implementing a FixAllProvider. However, I quickly realized that `AvoidUsingRedundantElseFixer.GetFixAllProvider` was never called during testing... thus the proposed fix.

Comment (meziantou) [reply src/Meziantou.Analyzer/Rules/AvoidUsingRedundantElseAnalyzer.cs]: I've created a PR (#150) to support batch fixer. It's not tested thoroughly but I think you can integrate it in your branch if you want to test the batch fixer

Comment (louis-z) [reply src/Meziantou.Analyzer/Rules/AvoidUsingRedundantElseAnalyzer.cs]: Ok, so based on your batch-fixer branch here's a new implementation.

File: src/Meziantou.Analyzer/Rules/AvoidUsingRedundantElseFixer.cs [hunk 2]
```diff
@@ -19,22 +19,39 @@ public sealed class AvoidUsingRedundantElseFixer : CodeFixProvider
...
                 equivalenceKey: title);
 
             context.RegisterCodeFix(codeAction, context.Diagnostics);
+
+            return Task.CompletedTask;
+        }
+
+        internal static async Task<Document> RemoveRedundantElseClausesInDocument(Document document, ImmutableArray<Diagnostic> diagnostics, CancellationToken cancellationToken)
+        {
+            foreach (var diagnostic in diagnostics.Reverse())
```
meziantou: Are you sure the diagnostic are sorted? Or should you use OrderByDescending(Location)?

tests/Meziantou.Analyzer.Test/Helpers/ProjectBuilder.Validation.cs
```diff
@@ -453,6 +475,34 @@ private static async Task<string> GetStringFromDocument(Document document)
...
+
+            public override async Task<IEnumerable<Diagnostic>> GetDocumentDiagnosticsAsync(Document document, CancellationToken cancellationToken)
+            {
+                var root = await document.GetSyntaxRootAsync().ConfigureAwait(false);
+                return _diagnostics.Where(d => root == d.Location.SourceTree.GetRoot());
+            }
+
+            public override Task<IEnumerable<Diagnostic>> GetProjectDiagnosticsAsync(Project project, CancellationToken cancellationToken)
+            {
+                // TODO using project.Documents
```
meziantou: While this is not needed in this PR, It would be great to implement it in a similar way as GetDocumentDiagnosticsAsync

tests/Meziantou.Analyzer.Test/Helpers/ProjectBuilder.Validation.cs
```diff
@@ -453,6 +475,34 @@ private static async Task<string> GetStringFromDocument(Document document)
...
+            private readonly Diagnostic[] _diagnostics;
+
+            public CustomDiagnosticProvider(Diagnostic[] diagnostics)
+            {
+                _diagnostics = diagnostics;
+            }
+
+            public override Task<IEnumerable<Diagnostic>> GetAllDiagnosticsAsync(Project project, CancellationToken cancellationToken)
+            {
+                // TODO I'm not sure of this one
```
meziantou: Now that you have implemented a FixAllProvider, do you know if this implementation is ok?

src/Meziantou.Analyzer/Rules/AvoidUsingRedundantElseAnalyzer.cs
meziantou: The implementation looks very good! I just add a few minor comments.

src/Meziantou.Analyzer/Rules/AvoidUsingRedundantElseFixer.cs
louis-z: You're right: I made an assumption, based on my test results.
Using OrderByDescending instead...

tests/Meziantou.Analyzer.Test/Helpers/ProjectBuilder.Validation.cs
louis-z: I'll have a look...

louis-z: I made a few changes. Let me know what you think.

meziantou: I'm not sure it is valid to check the `FilePath`. As indicated [in the comment](https://github.com/dotnet/roslyn/blob/2869fb77525e6128243e7607760c2dbe75864721/src/Compilers/Core/Portable/Syntax/SyntaxTree.cs#L38), the path can be empty if the source is generated and the value may not be unique. It may be safer to compare the syntaxtree or syntaxroot.

louis-z: Despite the Roslyn comment, I felt that using `FilePath` here was not a bad idea, because:
1. We control what gets stored in the `Document` objects we create (and I made sure we stored a non-empty FilePath)
1. [StyleCop also uses FilePath](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/ce8ee2f198f6138b04b968ffb73e1dc6d6e66686/StyleCop.Analyzers/StyleCopTester/TesterDiagnosticProvider.cs#L51) in their tests to determine whether a `Diagnostic` corresponds to a given `Document`. Note, however, that they make a distinction between `Project`-specific Diagnostics and `Document`-specific ones.

Nevertheless, I can go back to using the SyntaxTree root.

Issue #147, opened 2020-04-09
MA0071 fixer generates invalid code


meziantou:
Case 1:

````c#
if (a==b)
{
    var toto = "";
    return;
}
else // Should not remove this else
{
    var toto = 10;
}
````

Case 2:

````c#
if (a==b)
{
    var toto = "";
    return;
}
else // Should not remove this else as it would change the scope of the using statement. It should be find with block (with brackets) using() { }
{
    using var toto = new MyDisposableType();
}
````

Case 3:

````c#
if (a==b)
{
    A();
    return;

    void A()
    {
    }
}
else // Should not remove this else as it will not compile
{
    A();
    void A()
    {
    }
}
````

louis-z:
You can assign this issue to me. I've already come up with an initial fix. I may refine it a bit before proposing it though.

PR #151, opened 2020-04-12
MA0071 rule now handles "using statement" and conflicting local declarations

louis-z:
- Fix for https://github.com/meziantou/Meziantou.Analyzer/issues/147
- Also, rename local variables in AvoidUsingRedundantElseAnalyzer & Fixer, to simplify them or clarify their purpose. For instance, rename the somewhat vague 'statementOrBlock' (representing an if statement's single or composite statement child) to 'thenStatement', inspired by the C# language reference documentation:
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/if-else

### DIFF
--- a/src/Meziantou.Analyzer/Rules/AvoidUsingRedundantElseAnalyzer.cs
+++ b/src/Meziantou.Analyzer/Rules/AvoidUsingRedundantElseAnalyzer.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Immutable;
+﻿using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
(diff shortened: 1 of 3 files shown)


src/Meziantou.Analyzer/Rules/AvoidUsingRedundantElseAnalyzer.cs
```diff
@@ -32,27 +33,64 @@ public override void Initialize(AnalysisContext context)
...
+            return elseClauseSyntax.Statement is BlockSyntax elseBlockSyntax ?
+                elseBlockSyntax.ChildNodes() :
+                new[] { elseClauseSyntax.Statement };
+        }
+
+        private static IEnumerable<string> FindLocalIdentifiersIn(SyntaxNode node)
+        {
+            return node
+                .DescendantNodes()
+                .Where(node => node.IsKind(SyntaxKind.VariableDeclarator) || node.IsKind(SyntaxKind.LocalFunctionStatement))
```
meziantou: Does it detect the following pattern?

````
if (a is bool b) // Is 'b' a VariableDeclarator
````

louis-z: Now it does. :)

Issue #88, opened 2019-08-11
New rule: Replace Select(o => (Sample)o) by Cast<Sample>()


meziantou:
````
details.Select(o => (Sample)o);
details.Cast<Sample>();
````

louis-z:
Shouldn't this issue now be closed?

meziantou:
It should be closed thanks to your PR :)

meziantou:
The results vary depending on the .NET version and platform. But the code of `Select` has been improved in the recent versions. You can see more iterators based on the actual collection type to reduce allocations and the need for a state machine (`yield return`). So, it is not surprising that `Select` could be faster in some scenarios.

PR #158, opened 2020-04-17
Implement the UseCastInsteadOfSelect analyzer & fixer

louis-z:
Fix for https://github.com/meziantou/Meziantou.Analyzer/issues/88

### DIFF
--- a/src/Meziantou.Analyzer/Internals/ContextExtensions.cs
+++ b/src/Meziantou.Analyzer/Internals/ContextExtensions.cs
@@ -1,5 +1,4 @@
-﻿using System;
-using System.Collections.Immutable;
+﻿using System.Collections.Immutable;
 using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
@@ -117,13 +116,6 @@ namespace Meziantou.Analyzer
             {
                 context.ReportDiagnostic(CreateDiagnostic(descriptor, location, properties, messageArgs));
             }
-        }
-
-        [Flags]
-        public enum DiagnosticReportOptions
-        {
-            None = 0x0,
-            ReportOnMethodName = 0x1
         }
     }
 }
--- a/src/Meziantou.Analyzer/Internals/ContextExtensions.cs
+++ b/src/Meziantou.Analyzer/Internals/ContextExtensions.cs
@@ -1,6 +1,10 @@
-﻿using System.Collections.Immutable;
+﻿using System;
+using System.Collections.Immutable;
+using System.Linq;
 using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
 
 namespace Meziantou.Analyzer
 {
@@ -77,6 +81,18 @@ namespace Meziantou.Analyzer
             context.ReportDiagnostic(CreateDiagnostic(descriptor, operation.Syntax.GetLocation(), properties, messageArgs));
         }
 
+        public static void ReportDiagnostic(this OperationAnalysisContext context, DiagnosticDescriptor descriptor, ImmutableDictionary<string, string> properties, IInvocationOperation operation, DiagnosticReportOptions options, params string[] messageArgs)
+        {
+            if (options.HasFlag(DiagnosticReportOptions.ReportOnMethodName) &&
+                operation.Syntax.ChildNodes().FirstOrDefault() is MemberAccessExpressionSyntax memberAccessExpression)
+            {
+                context.ReportDiagnostic(Diagnostic.Create(descriptor, memberAccessExpression.Name.GetLocation(), properties, messageArgs));
+                return;
+            }
+
+            context.ReportDiagnostic(descriptor, properties, operation, messageArgs);
+        }
+
         public static void ReportDiagnostic(this CompilationAnalysisContext context, DiagnosticDescriptor descriptor, ISymbol symbol, params string[] messageArgs)
         {
             ReportDiagnostic(context, descriptor, ImmutableDictionary<string, string>.Empty, symbol, messageArgs);
@@ -101,6 +117,13 @@ namespace Meziantou.Analyzer
             {
                 context.ReportDiagnostic(CreateDiagnostic(descriptor, location, properties, messageArgs));
             }
+        }
+
+        [Flags]
+        public enum DiagnosticReportOptions
+        {
+            None = 0x0,
+            ReportOnMethodName = 0x1
         }
     }
 }
--- a/src/Meziantou.Analyzer/Rules/OptimizeLinqUsageAnalyzer.cs
+++ b/src/Meziantou.Analyzer/Rules/OptimizeLinqUsageAnalyzer.cs
@@ -610,23 +610,12 @@ namespace Meziantou.Analyzer.Rules
             if (castOp.Operand.Kind != OperationKind.ParameterReference)
                 return;
 
-            // Determine if we're casting to a nullable type. If we can't do it via Syntax, use ISymbol.
+            // Determine if we're casting to a nullable type.
             // TODO: Revisit this once https://github.com/dotnet/roslyn/pull/42403 is merged.
-            NullableFlowState nullableFlowState = NullableFlowState.None;
-            if (castOp.Syntax is CastExpressionSyntax castExpression)
-            {
-                if (castExpression.Type.IsKind(SyntaxKind.NullableType))
-                    nullableFlowState = NullableFlowState.MaybeNull;
-            }
-            else
-            {
-                var castMethodSymbol = operation.SemanticModel.GetSymbolInfo(castOp.Syntax.Parent).Symbol as IMethodSymbol;
-                var returnTypeSymbol = castMethodSymbol?.ReturnType;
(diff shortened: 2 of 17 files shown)


src/Meziantou.Analyzer/Rules/OptimizeLinqUsageAnalyzer.cs
```diff
@@ -564,6 +582,64 @@ bool HasTake()
...
+            if (!(returnOp.ReturnedValue is IConversionOperation castOp))
+                return;
+
+            // And is the cast applied to the sequence element, passed in as argument to the selector?
+            if (!(castOp.Operand is IParameterReferenceOperation))
+                return;
+
+            string castType = null;
+
+            // In C# there's 2 ways of casting...
```
meziantou: I'm not sure to understand why this is important. I think you should only use the properties of `IConversionOperation` (IsTryCast, conversion type).

tests/Meziantou.Analyzer.Test/Rules/OptimizeLinqUsageAnalyzerUseCastInsteadOfSelectTests.cs
```diff
@@ -0,0 +1,80 @@
...
+        {
+            return new ProjectBuilder()
+                .WithAnalyzer<OptimizeLinqUsageAnalyzer>(id: RuleIdentifiers.OptimizeEnumerable_CastInsteadOfSelect)
+                .WithCodeFixProvider<OptimizeLinqUsageFixer>();
+        }
+
+        [Theory]
+        [InlineData("enumerable.[|Select|](dt => (BaseType)dt)",
+                    "enumerable.Cast<BaseType>()")]
+        [InlineData("Enumerable.[|Select|](enumerable, dt => dt as Test.BaseType).Where(x => x != null)",
```
meziantou: The fix doesn't have the same behavior as the original code.

````
items.Select(item => item as Test.BaseType).Where(x => x != null) // Doesn't throw if an item is not a BaseType
items.OfType<BaseType>()  // Equivalent
items.Cast<BaseType>().Where(x => x != null) // Not equivalent, throw if an item is not a BaseType (InvalidCastException)
````

You should add a test to validate the `as` is not reported.

File: tests/Meziantou.Analyzer.Test/Rules/OptimizeLinqUsageAnalyzerUseCastInsteadOfSelectTests.cs [hunk 3]
```diff
@@ -0,0 +1,80 @@
+﻿using System.Threading.Tasks;
```
meziantou, line 1: I think you should add a test like the following

````
Enumerable.Range(0, 1).Select<int, object>(i=>i); // There is an implicit conversion but using Cast is valid here
Enumerable.Range(0, 1).Cast<object>()
````

File: src/Meziantou.Analyzer/Rules/OptimizeLinqUsageAnalyzer.cs [hunk 4]
```diff
@@ -564,6 +582,64 @@ bool HasTake()
...
+            {
+                return;
+            }
+
+            // Store the exact syntax of the type specification. The fixer will use it as is
+            // (without attempting simplification) when replacing Select by Cast.
+            var properties = CreateProperties(OptimizeLinqUsageData.UseCastInsteadOfSelect)
+               .Add("CastType", castType);
+
+            // Get the location of the [|Select|] member access expression
```
meziantou: I think you can get it using operations instead of c# syntax node.

tests/Meziantou.Analyzer.Test/Rules/OptimizeLinqUsageAnalyzerUseCastInsteadOfSelectTests.cs
louis-z: You're right, my bad.

Modified the tests and adapted code so that using `as` operator will no longer trigger a diagnostic.

We could also decide to simply offer another fix for it: the use of `OfType`. What do you think?

meziantou: `OfType<T>()` is the equivalent of `Select(item => item as Test.BaseType).Where(x => x != null)`. This is more complex to detect, but if you want to, you can create another PR to support it.

src/Meziantou.Analyzer/Rules/OptimizeLinqUsageAnalyzer.cs
```diff
@@ -564,6 +582,53 @@ bool HasTake()
             }
         }
 
+        private static void UseCastInsteadOfSelect(OperationAnalysisContext context, IInvocationOperation operation)
+        {
+            if (!string.Equals(operation.TargetMethod.Name, nameof(Enumerable.Select), StringComparison.Ordinal))
```
meziantou, line 587: You should also check that the TargetMethod'scontaining type is `System.Linq.Enumerable`.

src/Meziantou.Analyzer/Rules/OptimizeLinqUsageAnalyzer.cs
```diff
@@ -564,6 +582,53 @@ bool HasTake()
...
+
+            // Is it returning a cast value?
+            if (!(returnOp.ReturnedValue is IConversionOperation castOp))
+                return;
+
+            // And is the cast applied to the sequence element, passed in as argument to the selector?
+            if (!(castOp.Operand is IParameterReferenceOperation))
+                return;
+
+            if (!(castOp.Syntax is CastExpressionSyntax castExpression))
```
meziantou: I think you can use `!castOp.IsTryCast` instead of using the C# syntax node.

src/Meziantou.Analyzer/Rules/OptimizeLinqUsageAnalyzer.cs
```diff
@@ -564,6 +582,53 @@ bool HasTake()
...
+                return;
+
+            // Get the location of the [|Select|] member access expression
+            if (!(operation.Syntax.ChildNodes().FirstOrDefault() is MemberAccessExpressionSyntax memberAccessExpression))
+                return;
+
+            var castType = castExpression.Type.ToString();
```
meziantou: I think this is equivalent to `castOp.Type`.

tests/Meziantou.Analyzer.Test/Rules/OptimizeLinqUsageAnalyzerUseCastInsteadOfSelectTests.cs
louis-z: Then I guess it's out of scope for now.

src/Meziantou.Analyzer/Rules/OptimizeLinqUsageAnalyzer.cs
louis-z: There's no need, since it is already checked beforehand:
https://github.com/meziantou/Meziantou.Analyzer/blob/81320b9a6d0a8691493414cffb5e9e38500d03bd/src/Meziantou.Analyzer/Rules/OptimizeLinqUsageAnalyzer.cs#L107

src/Meziantou.Analyzer/Rules/OptimizeLinqUsageAnalyzer.cs
louis-z: I opted for
 `castOp.Type.ToMinimalDisplayString(operation.SemanticModel, operation.Syntax.SpanStart);`
instead.

This gives me, for instance, `object` rather than `Object`. It also keeps to a minimum the name qualification.

src/Meziantou.Analyzer/Rules/OptimizeLinqUsageAnalyzer.cs
louis-z: **Context**

Other LINQ-related rules highlight the `source` sequence onto which LINQ methods are called:
image

However, I wanted the new rule to highlight the `Select` member access expression that actually triggered the diagnostic.

**Implementation**

AFAIK the `IOperation` tree never goes down to the Identifier Name level; the location of the `Select` member access expression can therefore only be retrieved via the SyntaxTree. Right?

meziantou: Ok, this is a good idea that we should also apply to other diagnostics :)
I think a better implementation would be to create a new method in `ContextExtensions` so it is reusable.

````c#
internal static class ContextExtensions
{
    public static void ReportDiagnostic(this OperationAnalysisContext context, DiagnosticDescriptor descriptor, IInvocationOperation operation, InvocationReportOptions options, params string[] messageArgs);
}

[Flags]
public enum InvocationReportOptions
{
    None = 0x0,
    ReportOnMethodName = 0x1
}
````

What do you think of that?

Comment (louis-z) [reply src/Meziantou.Analyzer/Rules/OptimizeLinqUsageAnalyzer.cs]: Good idea. Just pushed a commit with those changes.

File: src/Meziantou.Analyzer/Internals/ContextExtensions.cs [hunk 8]
```diff
@@ -102,5 +118,12 @@ public static void ReportDiagnostic(this OperationBlockAnalysisContext context,
                 context.ReportDiagnostic(CreateDiagnostic(descriptor, location, properties, messageArgs));
             }
         }
+
+        [Flags]
+        public enum DiagnosticReportOptions
```
meziantou: Move it to another file

PR #177, opened 2020-08-10
Don't report MA0078 when the cast is required

### LINKED ISSUE #176: MA0078 (Cast instead of Select) should not apply when non-value-preserving conversions or custom operators are in play
(it should continue to offer this when there is a value-preserving conversion, for example an `enum : byte` => `byte`)

Otherwise you can get:

> System.InvalidCastException: Unable to cast object of type '{blah}' to type '{whatever}'.
>    at System.Linq.Enumerable.CastIterator[TResult](IEnumerable source)+MoveNext()

Simple example (non-value-preserving conversion):

``` c#
int[] values = {1, 2, 3};
// suggests MA0078 use 'Cast<int>' instead of 'Select' to cast
var arr1 = values.Select(i => (byte)i).ToArray(); // works

var arr2 = values.Cast<byte>().ToArray(); // fails
```

More nuanced example (custom operator):

``` c#

using System;
using System.Linq;

static class P
{
    static void Main()
    {
        var foos = new[] { new Foo("1"), new Foo("42") };
        Console.WriteLine("Testing " + nameof(Enumerable.Select));
        try
        {
            // suggests MA0078 use 'Cast<int>' instead of 'Select' to cast
            foreach (var i in foos.Select(x => (int)x))
            {
                Console.WriteLine(i);
            }
        }
        catch (Exception ex)
        {
            Console.WriteLine(ex);
        }

        // this fails with
        // System.InvalidCastException: Unable to cast object of type 'Foo' to type 'System.Int32'.
        Console.WriteLine("Testing " + nameof(Enumerable.Cast));
        try
        {
            foreach (var i in foos.Cast<int>())
            {
                Console.WriteLine(i);
            }
        }
        catch (Exception ex)
        {
            Console.WriteLine(ex);
        }
    }
}

class Foo
{
    private readonly string _value;
    public Foo(string value)
        => _value = value;

    public static implicit operator int(Foo foo) => int.Parse(foo._value, CultureInfo.InvariantCulture);
}
```

meziantou:
fix #176

### DIFF
--- a/src/Meziantou.Analyzer/Internals/TypeSymbolExtensions.cs
+++ b/src/Meziantou.Analyzer/Internals/TypeSymbolExtensions.cs
@@ -126,7 +126,12 @@ namespace Meziantou.Analyzer
 
         public static bool IsEnumeration(this ITypeSymbol? symbol)
         {
-            return symbol is INamedTypeSymbol namedTypeSymbol && namedTypeSymbol.EnumUnderlyingType != null;
+            return GetEnumerationType(symbol) != null;
+        }
+        
+        public static INamedTypeSymbol? GetEnumerationType(this ITypeSymbol? symbol)
+        {
+            return (symbol as INamedTypeSymbol)?.EnumUnderlyingType;
         }
 
         public static bool IsNumberType(this ITypeSymbol? symbol)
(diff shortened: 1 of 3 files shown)


mgravell:
Looking good! the enum one is still a little dangerous (wrong-sized enums is actually the main thing that bit me :p)

meziantou:
The last commit should handle enums correctly. Could you have a look before I merge and publish a new version of the analyzer?

PR #205, opened 2020-11-11
Support for Operators for MA0015

rsking:
MA0015 does not support operators, and will always fail. This PR adds support for operators so that the parameter names can be found.


meziantou:
Thanks for your contribution!

PR #227, opened 2021-03-22
Fix branch name from master to main

juchom:
Simple fix that avoid the "branch not found" message from github when clicking on rule link in Visual Studio


meziantou:
Good catch! Thank you 😊

meziantou:
Your change should be available in [Meziantou.Analyzer@1.0.648](https://www.nuget.org/packages/Meziantou.Analyzer/1.0.648)

Issue #230, opened 2021-03-26
There are no destructors in C#


quixoticaxis:
C# has finalizers with destructor syntax. Updating the wording for MA0055 and MA0086 would be great.

meziantou:
The .NET documentation use "Finalizers" while the spec use "Destructors". Both use "destructor" in their page url.
If you want to can submit a Pull Request to change destructor to finalizer.

- https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/destructors
- https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/classes#destructors

quixoticaxis:
@meziantou the language specification ([ECMA-334 5th Edition / December 2017](https://www.ecma-international.org/publications-and-standards/standards/ecma-334/)) actually uses the term "finalizer" and contains the following clause:
> 15.13 Finalizers
> [Note: In an earlier version of this standard, what is now referred to as a "finalizer" was called a 
> "destructor". Experience has shown that the term "destructor" caused confusion and often resulted to 
> incorrect expectations, especially to programmers knowing C++. In C++, a destructor is called in a 
> determinate manner, whereas, in C#, a finalizer is not. To get determinate behavior from C#, one should
> use Dispose. end note]

It seems that online documentation is out-of-sync. Not that it matters much, of course.

PR #231, opened 2021-03-27
Updated the descriptions of MA0055 and MA0086 to conform to the latest ECMA specification

quixoticaxis:
This PR follows the #230.

I've updated the descriptions, and renamed the constants/classes/containing files accordingly. The existing tests do pass.


meziantou:
Thanks!

PR #252, opened 2021-09-01
Fix MA0011 doc - swap DateTime.TryParse parameters

jannesrsa:
The **DateTimeStyles** and **IFormatProvider** parameters should be swapped.

https://docs.microsoft.com/en-us/dotnet/api/system.datetime.tryparse?view=net-5.0#System_DateTime_TryParse_System_String_System_IFormatProvider_System_Globalization_DateTimeStyles_System_DateTime__

````csharp
public static bool TryParse (string? s, IFormatProvider? provider, System.Globalization.DateTimeStyles styles, out DateTime result);
````


meziantou:
Thanks for improving the documentation 🚀

PR #298, opened 2022-02-04
Fix English grammar issues

louis-z:
- "Prefer" should be followed by a gerund
- A singular subject "that has" rather than "that have" something
- We dispose "of" something
- Plus, a bit more


meziantou:
Thanks!

PR #367, opened 2022-09-01
Document rationale and types that are special-cased.

### LINKED ISSUE #365: Consider not reporting MA0075 / MA0076 when IFormattable is implemented via explicit interface implementation
**Context**
Rule identifier: _MA0075_
Target Framework: _.NET 6.0_
Language version: _10.0_

**Background**
In .NET 6.0 (and perhaps earlier), **System.Version** implements **ISpanFormattable**. Since **ISpanFormattable** implement **IFormattable**, **System.Version** also had to implement `string IFormattable.ToString(string? format, IFormatProvider? formatProvider)`.

This was done via an explicit interface implementation:
```c#
string IFormattable.ToString(string? format, IFormatProvider? formatProvider) =>
            ToString();
```

Source is available [here](https://github.com/dotnet/runtime/blob/e98fc2a32fd0ea90092d5f93485e2aebd7ec0402/src/libraries/System.Private.CoreLib/src/System/Version.cs#L191-L192).

I suppose the reason why it was done this way, is because there's nothing culture specific in the `ToString()` implementation of **System.Version**. They probably only wanted to implement the `bool TryFormat (Span<char> destination, out int charsWritten, ReadOnlySpan<char> format, IFormatProvider? provider)` method of **ISpanFormattable**, but had no choice.

You could argue that, if a given class only makes available an `ToString(string format, IFormatProvider formatProvider)` method via an explicit interface implementation, that it does not want users to actually invoke it.

**Describe the bug**
When targeting .NET 6.0 or higher, **MA0075** / **MA0076** is reported for the following code fragment:

```c#
Console.WriteLine($"Running on {System.Environment.Version}");
Console.WriteLine("Running on " + System.Environment.Version);
```

There are two solutions:
* Special-case **System.Version** like you already do for other types.
* Do not report **MA0075** / **MA0076** when the type only implements **IFormattable** via an explicit interface implementation (meaning, it has no public `ToString(string format, IFormatProvider formatProvider)` method).

The latter would avoid reporting **MA0075** / **MA0076** for any type that only implements **IFormattable** via an explicit interface implementation, and not just those that are special-cased in [`DoNotUseImplicitCultureSensitiveToStringAnalyzer. IsFormattableType(Compilation compilation, ITypeSymbol typeSymbol)`](https://github.com/meziantou/Meziantou.Analyzer/blob/b6f1cd8721f87655e4a2ae6945155c826154ffbf/src/Meziantou.Analyzer/Rules/DoNotUseImplicitCultureSensitiveToStringAnalyzer.cs#L154-L193).

For example:
```c#
var address = new Address("xx", 50);
Console.WriteLine($"You live at {address}");
Console.WriteLine("You live at " + address);

private class Address : IFormattable
{
    public Address(string street, int houseNumber)
    {
        Street = street;
        HouseNumber = houseNumber;
    }

    public string Street { get; }

    public int HouseNumber { get; }

    public override string ToString()
    {
        return $"{Street}, {HouseNumber}";
    }

    string IFormattable.ToString(string format, IFormatProvider formatProvider)
    {
        return ToString();
    }
}
```

drieseng:
Document rationale and types that are special-cased.
Document for which types no diagnostic is reported if the operand is a positive constants.
No longer special case positive constants of type decimal, double and float.

Fixes #365.


drieseng:
@meziantou, perhaps consider adding [UInt128](https://docs.microsoft.com/en-us/dotnet/api/system.uint128?view=net-7.0) while we're at it?

meziantou:
Good idea! You can add UInt128 to the exclusion list

docs/Rules/MA0075.md
```diff
@@ -1,12 +1,28 @@
...
+This rule reports cases where an implicit call to `obj.ToString()` may differ based on the current culture:
+
 ````csharp
 _ = "abc" + -1; // non-compliant as the result depends on the current culture
 
 _ = "abc" + -1.ToString(CultureInfo.CurrentCulture);   // compliant
 _ = "abc" + -1.ToString(CultureInfo.InvariantCulture); // compliant
 ````
 
+A given type is considered to have a culture-sensitive `ToString()` implementation if it implements **System.IFormattable**.<br/>
```
meziantou:
```suggestion
A given type is considered to have a culture-sensitive `ToString()` implementation if it implements `System.IFormattable`.<br/>
```

docs/Rules/MA0075.md
```diff
@@ -1,12 +1,28 @@
...
+An exception is made for the following types:
```
meziantou, line 13: You can also specify that positive-numbers are excluded

drieseng: Hmm, not sure if that is a good idea (at least not for float and double):

```c#
const double value = 1.1d;

CultureInfo.CurrentCulture = new CultureInfo("en-US");
Console.WriteLine("" + value);

CultureInfo.CurrentCulture = new CultureInfo("nl-BE");
Console.WriteLine("" + value);
```

Result:

```txt
1.1
1,1
```

No problem, let's exclude fractional numbers then. Wait, hold on ...

```c#
const double value = 94343434232323323232d;

CultureInfo.CurrentCulture = new CultureInfo("en-US");
Console.WriteLine("" + value);

CultureInfo.CurrentCulture = new CultureInfo("nl-BE");
Console.WriteLine("" + value);
```

Result:

```txt
9.434343423232333E+19
9,434343423232333E+19
```

Perhaps there are even other possible issues with this. Could there, for example, be a locale that always shows a sign for positive numbers?

meziantou: I mean the analyzer already skips positive integer numbers (or look alike positive).

https://github.com/meziantou/Meziantou.Analyzer/blob/main/src/Meziantou.Analyzer/Rules/DoNotUseImplicitCultureSensitiveToStringAnalyzer.cs#L200

You are right on the floating numbers. I think we should update the code to remove the special case for Single/Double/Decimal.

meziantou: > Perhaps there are even other possible issues with this. Could there, for example, be a locale that always shows a sign for positive numbers?

The code doesn't use the culture at all for positive numbers (for Int32 and other integer values). You can check the implementation here: https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Int32.cs,106

meziantou:
Thanks!

PR #399, opened 2022-10-02
Add ReturnDirectlyAnalyzer

### LINKED ISSUE #391: Interest in a ReturnTaskDirectlyAnalyzer?
Hi, I made an analyzer which detects instances of a `Task` which can be returned directly: https://github.com/CollinAlpert/ReturnTaskDirectlyAnalyzer

I'd love more people to use it and was wondering if you'd be interested in me opening a PR adding it to this package.

CollinAlpert:
This PR adds an analyzer for detecting usages of the async/await pattern where the `Task` could be returned directly instead of awaiting it.
I'm opening as draft, since I had some problems adding the tests. You test structure doesn't seem to allow for multiple diagnostic locations. This is how I solved it in my project: https://github.com/CollinAlpert/ReturnTaskDirectlyAnalyzer/blob/master/ReturnTaskDirectlyAnalyzer.Tests/ReturnTaskDirectlyTests.cs#L835
Maybe you can help me integrate this in your structure.

Fixes #391.


meziantou:
Thanks for the PR! I'll review it this evening or tomorrow.

> Maybe you can help me integrate this in your structure.

The project is quite old, and the test framework provided by Roslyn was not available (or easy to use) back then. So, it is custom. If you want to validate multiple diagnostics in the same method, you can use the "span" declaration (quite similar to Roslyn)

````c#
var code = """
// 3 diagnostics are expected
public [|void|] Sample() // a diagnostic is expected on "void"
{
    [|int|] a = 0; // a diagnostic is expected on "int"
    [||]Demo.Sample(); // a diagnostic starting on "Demo" is expected
}
"""
````

If you want to validate the fixer, the common pattern looks like https://github.com/meziantou/Meziantou.Analyzer/blob/main/tests/Meziantou.Analyzer.Test/Rules/MakeClassStaticAnalyzerTests.cs#L77-L95.

File: tests/Meziantou.Analyzer.Test/Rules/ReturnTaskDirectlyTests.cs [hunk 1]
```diff
@@ -0,0 +1,895 @@
...
+using System.Threading.Tasks;
+using Meziantou.Analyzer.Rules;
+using TestHelper;
+using Xunit;
+
+namespace Meziantou.Analyzer.Test.Rules;
+
+public class ReturnTaskDirectlyTests
+{
+	private const string Scaffold = @"
```
meziantou: This project requires .NET 7, so you can use the new C# raw string syntax using `"""` (https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-11?WT.mc_id=DT-MVP-5003978#raw-string-literals)

tests/Meziantou.Analyzer.Test/Rules/ReturnTaskDirectlyTests.cs
```diff
@@ -0,0 +1,895 @@
...
+	{
+		var source = string.Format(CultureInfo.InvariantCulture, Scaffold, method);
+		var fixedSource = string.Format(CultureInfo.InvariantCulture, Scaffold, fixedMethod);
+
+        return new ProjectBuilder()
+            .WithAnalyzer<ReturnTaskDirectlyAnalyzer>()
+            .WithCodeFixProvider<ReturnTaskDirectlyFixer>()
+            .WithSourceCode(source)
+            .ShouldFixCodeWith(fixedSource)
+            .AddAsyncInterfaceApi()
```
meziantou: It easier to target a recent framework

```suggestion
            .WithTargetFramework(TargetFramework.Net7_0)
```

tests/Meziantou.Analyzer.Test/Rules/ReturnTaskDirectlyTests.cs
```diff
@@ -0,0 +1,895 @@
...
+using System.Threading.Tasks;
+
+namespace UnitTests;
+
+public class Test {{
+	private readonly DataService _dataService;
+
+	{0}
+
+	public Task DoSomethingAsync() {{
```
meziantou: I'm not sure you need all these methods in the tests. I think you can use _well-known_ methods from the BCL such as `Task.Delay(0)`, `Task.FromResult`, an so on. It makes the tests easier to read as we can quickly understand the code.

Also, there is less code to analyze, so it's much easier to debug

tests/Meziantou.Analyzer.Test/Rules/ReturnTaskDirectlyTests.cs
```diff
@@ -0,0 +1,895 @@
...
+	[InlineData(WithMultipleAwaits, WithMultipleAwaitsFixed, 2)]
+	[InlineData(WithLocalFunction, WithLocalFunctionFixed, 2)]
+	[InlineData(LocalFunction, LocalFunctionFixed)]
+	[InlineData(WithConfigureAwait, WithConfigureAwaitFixed, 2)]
+	[InlineData(WithUnrelatedUsingStatement, WithUnrelatedUsingStatementFixed)]
+	[InlineData(LambdaExpression, LambdaExpressionFixed)]
+	[InlineData(LambdaExpressionWithReturn, LambdaExpressionWithReturnFixed)]
+	[InlineData(LambdaBlock, LambdaBlockFixed, 2)]
+	[InlineData(LambdaBlockWithReturn, LambdaBlockWithReturnFixed, 2)]
+	public Task ShouldRaiseAsync(string method, string fixedMethod, int diagnosticLocations = 1)
```
meziantou: I prefer creating multiple tests instead of using `InlineData`. It is easier to find and run a specific test from VS.

If you look at other test files, you can see that I try to extract the initialization in a method:

````
        return new ProjectBuilder()
            .WithAnalyzer<ReturnTaskDirectlyAnalyzer>()
            .WithCodeFixProvider<ReturnTaskDirectlyFixer>()
````

File: tests/Meziantou.Analyzer.Test/Rules/ReturnTaskDirectlyTests.cs [hunk 5]
```diff
@@ -0,0 +1,895 @@
...
+	}}
+}}
+
+public class DataService {{
+	public Task<List<int>> GetListAsync() {{
+		return Task.FromResult(new List<int>() {{ 1, 2, 3 }});
+	}}
+}}";
+
+	// I know regions are bad, please don't report me to the region police.
```
meziantou: 👮 A better solution may be to split the tests in 2 different files (partial class if it is easier)

tests/Meziantou.Analyzer.Test/Rules/ReturnTaskDirectlyTests.cs
```diff
@@ -0,0 +1,895 @@
...
+public class Test {{
+	private readonly DataService _dataService;
+
+	{0}
+
+	public Task DoSomethingAsync() {{
+		return Task.Delay(1000);
+	}}
+
+	public Task<int> GetSomethingAsync() {{
```
meziantou: Most of the time, the implementation of helper methods is not important. You can replace the implementation with `throw null`. It is easier to see what is not important in the test.

````
public Task<int> GetSomethingAsync() => throw null;
````

File: src/Meziantou.Analyzer/Rules/ReturnTaskDirectlyAnalyzer.cs [hunk 7]
```diff
@@ -0,0 +1,215 @@
...
+    public override void Initialize(AnalysisContext context)
+	{
+		context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+		context.EnableConcurrentExecution();
+		context.RegisterCompilationStartAction(OnCompilationStart);
+	}
+
+	private static void OnCompilationStart(CompilationStartAnalysisContext context)
+	{
+		context.RegisterSyntaxNodeAction(OnMethodAnalysis, SyntaxKind.MethodDeclaration, SyntaxKind.LocalFunctionStatement, SyntaxKind.ParenthesizedLambdaExpression);
```
meziantou: You can create a _context_ to store all symbols (TaskSymbol, etc.). This way, you only need to find them once per compilation which is faster. You can find samples in other rules such as https://github.com/meziantou/Meziantou.Analyzer/blob/main/src/Meziantou.Analyzer/Rules/AwaitTaskBeforeDisposingResourcesAnalyzer.cs#L30-L59.

src/Meziantou.Analyzer/Rules/ReturnTaskDirectlyAnalyzer.cs
```diff
@@ -0,0 +1,215 @@
...
+		}
+		else if (context.Node.IsKind(SyntaxKind.ParenthesizedLambdaExpression))
+		{
+			var lambdaSymbol = (IMethodSymbol?)context.SemanticModel.GetSymbolInfo(node, cancellationToken: context.CancellationToken).Symbol;
+			if (!IsMethodValid(node, lambdaSymbol))
+			{
+				return;
+			}
+
+			body = (node as ParenthesizedLambdaExpressionSyntax)!.Block;
```
meziantou: Use an explicit cast

```suggestion
			body = ((ParenthesizedLambdaExpressionSyntax)node).Block;
```

src/Meziantou.Analyzer/Rules/ReturnTaskDirectlyAnalyzer.cs
meziantou: Have you tried using the Operation analysis instead of SyntaxNode analysis? Most of the time, this simplifies the code as you don't have to deal with the different ways to declare the same concept. Also, you don't need to use the SemanticModel as the main info as already accessible from operations.

tests/Meziantou.Analyzer.Test/Rules/ReturnTaskDirectlyTests.cs
CollinAlpert: Oh sweet! Done.

tests/Meziantou.Analyzer.Test/Rules/ReturnTaskDirectlyTests.cs
CollinAlpert: Done.

tests/Meziantou.Analyzer.Test/Rules/ReturnTaskDirectlyTests.cs
CollinAlpert: Done.

tests/Meziantou.Analyzer.Test/Rules/ReturnTaskDirectlyTests.cs
CollinAlpert: Done.

tests/Meziantou.Analyzer.Test/Rules/ReturnTaskDirectlyTests.cs
CollinAlpert: Done.

tests/Meziantou.Analyzer.Test/Rules/ReturnTaskDirectlyTests.cs
CollinAlpert: Done.

src/Meziantou.Analyzer/Rules/ReturnTaskDirectlyAnalyzer.cs
CollinAlpert: Good idea! Done.

src/Meziantou.Analyzer/Rules/ReturnTaskDirectlyAnalyzer.cs
CollinAlpert: Done.

src/Meziantou.Analyzer/Rules/ReturnTaskDirectlyAnalyzer.cs
CollinAlpert: Yeah, I was thinking that as well as I kept adding cases to the analyzer. I'll see if I can find an elegant solution.

meziantou:
> Still 7 tests failing, which is better than before :) The test framework doesn't seem to recognize the additional locations which the analyzer emits.

Indeed, additional locations are not supported. I don't think this is very important when the diagnostic provides a code fix. You can validate the main `Location` in the test. And if you validate the code fix, other locations should be implicitly tested as all locations should be updated.

That's being said, I'll see if it's easy to add as it could be useful.

src/Meziantou.Analyzer/Rules/ReturnTaskDirectlyAnalyzer.cs
CollinAlpert: Hmm, this doesn't seem as straightforward as I thought. Access to method modifiers and return types is easier with a `SyntaxNode` and also there doesn't seem to be an `IOperation` for a lambda expression, so I would need to process that as a `SyntaxNode` anyway.

meziantou:
> That's being said, I'll see if it's easy to add as it could be useful.

That's not as easy as I thought. I'm not sure I want to add this as the best move would be to adopt the test framework from Roslyn (even if this is a tedious change).

At the moment, you can make public the method `ShouldReportDiagnostic`. It is harder than the span syntax to use, but it should work.

src/Meziantou.Analyzer/Rules/ReturnTaskDirectlyAnalyzer.cs
```diff
@@ -0,0 +1,233 @@
...
+	{
+		context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+		context.EnableConcurrentExecution();
+		context.RegisterCompilationStartAction(OnCompilationStart);
+	}
+
+	private static void OnCompilationStart(CompilationStartAnalysisContext context)
+    {
+        var analyzerContext = new AnalyzerContext(context.Compilation);
+        context.RegisterSyntaxNodeAction(ctx => OnMethodAnalysis(ctx, analyzerContext), SyntaxKind.MethodDeclaration, SyntaxKind.LocalFunctionStatement, SyntaxKind.ParenthesizedLambdaExpression);
```
meziantou: You can move `OnMethodAnalysis` into the `AnalyzerContext` class and remove the `static` modifier. It will simplify the code as you can use the properties directly.

```suggestion
        context.RegisterSyntaxNodeAction(ctx => analyzerContext.OnMethodAnalysis(ctx), SyntaxKind.MethodDeclaration, SyntaxKind.LocalFunctionStatement, SyntaxKind.ParenthesizedLambdaExpression);
```

src/Meziantou.Analyzer/Rules/ReturnTaskDirectlyAnalyzer.cs
meziantou: > there doesn't seem to be an IOperation for a lambda expression

You can use `IDelegateCreationOperation` or `OperationKind.DelegateCreation`

> Access to method modifiers and return types is easier with a `SyntaxNode`

I don't think you need to access modifiers. Your code already use properties from IMethodSymbol.


````c#
context.RegisterOperationBlockAction(ctx => { var block = ctx.OperationBlocks; var method = ctx.OwningSymbol; });
context.RegisterOperationAction(ctx => { var op = ctx.Operation; }, OperationKind.LocalFunction);
context.RegisterOperationAction(ctx => { var op = ctx.Operation; }, OperationKind.DelegateCreation);
````

meziantou:
@CollinAlpert I don't know if you still work on this PR. Let me know if you want me to finish the tests and merge it.

CollinAlpert:
I have continued working here and there, however adding the operation analysis has been harder than thought. Would it be okay to merge the PR with syntax analysis?

meziantou:
> Would it be okay to merge the PR with syntax analysis?

Yes, if it is too hard to use Operations, I'm fine with using the syntax analysis.

Issue #401, opened 2022-10-13
Diagnostic for maximum statements reads "nn lines" instead of "nn statements"


f3:
**Context**
Rule identifier: _MA0051_
Target Framework: all
Language version: _any_

**Describe the bug**
When the number of **statements** is exceeded for a method, the Diagnostic for `MA0051` outputs:

> warning MA0051: Method is too long (41 lines; maximum allowed: 40)

This can be confusing if the environment has been set up to ignore  `MA0051` for maximum number of _lines_.  It should probably read:

> warning MA0051: Method is too long (41 **_statements_**; maximum allowed: 40)

**To Reproduce**
Code or steps to reproduce the behavior:

Set up `.editorconfig` to ignore maximum lines, but configure maximum _statements_ to something ridiculous 😁

````text
MA0051.maximum_lines_per_method = -1
MA0051.maximum_statements_per_method = 10
````
write some code having more than 10 statements in it.

**Additional context**

[Relevant line of code](https://github.com/meziantou/Meziantou.Analyzer/blob/main/src/Meziantou.Analyzer/Rules/MethodShouldNotBeTooLongAnalyzer.cs#L100)

PR #402, opened 2022-10-13
[#401] Modify method-too-long output for statement count

f3:
When the number of statements is exceeded for a method, the Diagnostic for `MA0051` outputs:

> warning MA0051: Method is too long (41 lines; maximum allowed: 40)

This can be confusing if the environment has been set up to ignore `MA0051` for maximum number of _lines_. The output should probably read:

> warning MA0051: Method is too long (41 statements; maximum allowed: 40)

Addresses #401


meziantou:
Thanks for your contribution 🚀

PR #410, opened 2022-10-28
Update ClassMustBeSealedAnalyzer.cs

nnpcYvIVl:
Spelling fix

### DIFF
--- a/src/Meziantou.Analyzer/Rules/ClassMustBeSealedAnalyzer.cs
+++ b/src/Meziantou.Analyzer/Rules/ClassMustBeSealedAnalyzer.cs
@@ -133,7 +133,7 @@ public sealed class ClassMustBeSealedAnalyzer : DiagnosticAnalyzer
 
         private static bool SealedClassWithVirtualMember(AnalyzerOptions options, ISymbol symbol)
         {
-            return options.GetConfigurationValue(symbol, RuleIdentifiers.ClassMustBeSealed + ".class_with_virtual_member_shoud_be_sealed", defaultValue: false);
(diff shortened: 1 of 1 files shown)


src/Meziantou.Analyzer/Rules/ClassMustBeSealedAnalyzer.cs
```diff
@@ -133,7 +133,7 @@ private static bool PublicClassShouldBeSealed(AnalyzerOptions options, ISymbol s
 
         private static bool SealedClassWithVirtualMember(AnalyzerOptions options, ISymbol symbol)
         {
-            return options.GetConfigurationValue(symbol, RuleIdentifiers.ClassMustBeSealed + ".class_with_virtual_member_shoud_be_sealed", defaultValue: false);
+            return options.GetConfigurationValue(symbol, RuleIdentifiers.ClassMustBeSealed + ".class_with_virtual_member_should_be_sealed", defaultValue: false);
```
meziantou, line 136: - This is a breaking change, and the option is publicly documented. So, you need to support the new name and the old name

```suggestion
            var defaultValue = options.GetConfigurationValue(symbol, RuleIdentifiers.ClassMustBeSealed + ".class_with_virtual_member_shoud_be_sealed", defaultValue: false);
            return options.GetConfigurationValue(symbol, RuleIdentifiers.ClassMustBeSealed + ".class_with_virtual_member_should_be_sealed", defaultValue);
```

- Could you update the tests in the same PR? Keep the existing test, and add a new one with the new option name to ensure both options are accepted.

meziantou (changes requested):

PR #518, opened 2023-05-05
Remove protected constructor check from MA130

jontdelorme:
GetType analyzer is incorrectly reporting error about protected constructor on abstract classes


meziantou:
Could you provide an example of a false-positive and add a test?

jontdelorme:
In one of our codebase, we had it disabled so the error was quite confusing since there were no GetType() call in the file

Issue #538, opened 2023-05-29
MA0042 is also reported for non-async methods


drieseng:
### Version of the Meziantou.Analyzer NuGet package

2.0.54

### Rule Identifier

MA0042

### Target Framework

net7.0

### C# Language version

C# 11

### Description

The doc page for **MA0042** makes me believe that it will not report a diagnostic for a method without **async** modifier:

> The rule only reports a diagnostic when the calling method is already async.

However, **MA0042** is reported when you have a using statement and the instance implements **IAsyncDisposable**.

### Reproduction Steps

Minimal code:

```c#
public static Task RunNotAsync(Resource resource)
{
    // This results in MA0042: Prefer using 'await using'
    using (resource)
    {
    }

    return Task.CompletedTask;
}

public sealed class Resource : IDisposable, IAsyncDisposable
{
    public void Dispose()
    {
    }

    public ValueTask DisposeAsync()
    {
        return ValueTask.CompletedTask;
    }
}
```

meziantou:
MA0042 is reported when the enclosing method has the `async` keyword or when the return type is awaitable (`Task`, `ValueTask`, `IAsyncEnumerable`, etc.). In this case, you can use the `await` keyword without changing the signature (adding `async` is an implementation detail, not a breaking change). So, in your example, I think the MA0042 should be reported as you can add the `async` keyword without changing the signature.

Maybe the documentation should be clarified.

Side-note: I recently changed the rule to detect awaitable types instead of hardcoding `Task` and `ValueTask`. But, I think I made a mistake. Instead of detecting awaitable types, it should detect types that support the `async` keyword (`Task`, `Task<T>`, `IAsyncEnumerable<T>`, `IAsyncEnumerator<T>`, types decorated with `[AsyncMethodBuilder]`)

PR #539, opened 2023-05-29
MA0042 correctly determine when a method can use async

meziantou:
Fix https://github.com/meziantou/Meziantou.Analyzer/issues/538

### DIFF
--- a/docs/Rules/MA0042.md
+++ b/docs/Rules/MA0042.md
@@ -14,7 +14,22 @@ public async Task Sample()
 }
 ````
 
-The rule only reports a diagnostic when the calling method is already `async`. [MA0045](MA0045.md) detects the same cases, but reports them even if applying a fix would require you to change the calling method's signature.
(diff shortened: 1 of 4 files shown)


docs/Rules/MA0042.md
```diff
@@ -14,7 +14,22 @@ public async Task Sample()
 }
 ````
 
-The rule only reports a diagnostic when the calling method is already `async`. [MA0045](MA0045.md) detects the same cases, but reports them even if applying a fix would require you to change the calling method's signature.
+The rule only reports a diagnostic when the enclosing method is already `async` or when the `async` keyword can be added with changing the return type. While you can use `async void`, the rule will not report a diagnostic when the method returns `void`. [MA0045](MA0045.md) detects the same cases, but reports them even if applying a fix would require you to change the calling method's signature.
+
+
```
meziantou:
```suggestion
```

PR #546, opened 2023-06-06
MA0076 should not report for TimeSpan without format

meziantou:


src/Meziantou.Analyzer/Internals/CultureSensitiveFormattingContext.cs
```diff
@@ -295,7 +295,7 @@ private static bool IsInvariantDateTimeFormat(IOperation? valueOperation)
 
     private static bool IsInvariantTimeSpanFormat(IOperation? valueOperation)
     {
-        return valueOperation is { ConstantValue: { HasValue: true, Value: "c" or "C" } };
+        return valueOperation == null || valueOperation is { ConstantValue: { HasValue: true, Value: null or "" or "c" or "C" } };
```
0liver, line 298: You're amazing! That was probably the fastest bug fix in open source software ever! 🏆

I'd like to draw your attention to the fact, that both `"t"` and `"T"` are also acceptable standard format strings with [the same effect as `"c"`](https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-timespan-format-strings#the-constant-c-format-specifier), so you might want to add those as well:

image


Thanks for delivering a great piece of software with even greater support!

meziantou: Added in https://github.com/meziantou/Meziantou.Analyzer/commit/715447944a8469484cd1b02e64301bd7e2f5bda7

Issue #597, opened 2023-09-08
MA0134 not triggered by synchronous Task-returning local method in async method


jackalvrus:
### Version of the Meziantou.Analyzer NuGet package

2.0.84

### Rule Identifier

MA0134

### Target Framework

net6.0

### C# Language version

C# 10

### Description

In our production code, I ran into a scenario where a synchronous task-returning but non-async local method inside an async method is not detected by MA0134. Adding a test confirmed that it was not being detected. 

### Reproduction Steps

This test fails:

```c#
    [Fact]
    public async Task Report_TaskInLocalTaskReturningFunction()
    {
        await CreateProjectBuilder()
            .WithSourceCode("""
                using System.Threading.Tasks;
                class Test
                {
                    async Task A()
                    {
                        await B();

                        Task B()
                        {
                            [||]Task.Delay(0);
                            return Task.CompletedTask;
                        }
                    }
                }
                """)
            .ValidateAsync();
    }
```

### Other information

Simply removing the early break-out for async methods in the lambda passed to `RegisterSymbolStartAction` seems to fix the issue.

PR #598, opened 2023-09-08
MA0134 is correctly reported in local function

jackalvrus:


jackalvrus:
Fixes #597

src/Meziantou.Analyzer/Rules/AwaitAwaitableMethodInSyncMethodAnalyzer.cs
```diff
@@ -30,9 +30,6 @@ public override void Initialize(AnalysisContext context)
             var awaitableTypes = new AwaitableTypes(context.Compilation);
             context.RegisterSymbolStartAction(context =>
             {
-                if (context.Symbol is IMethodSymbol method && (method.IsAsync || method.IsTopLevelStatementsEntryPointMethod()))
-                    return; // Already handled by CS4014
```
meziantou, line 34: Can you move the comment `// Already handled by CS4014` to the line 58?

Also, you can remove the `// While there is a check in RegisterSymbolStartAction, this is needed to handle lambda and delegates` comment as it is not valid anymore.

jackalvrus: Good catch. Comments fixed.

meziantou:
Thanks! The new version (2.0.85) will be available soon :)

PR #738, opened 2024-06-26
disable MA0016, if the class is decorated with SerializableAttribute

Micke3rd:
with that change rule MA0016 is disabled, if the class is decorated with SerializableAttribute

### DIFF
--- a/src/Meziantou.Analyzer/Rules/PreferReturningCollectionAbstractionInsteadOfImplementationAnalyzer.cs
+++ b/src/Meziantou.Analyzer/Rules/PreferReturningCollectionAbstractionInsteadOfImplementationAnalyzer.cs
@@ -49,6 +49,7 @@ public sealed class PreferReturningCollectionAbstractionInsteadOfImplementationA
 
             XmlClassAttributeSymbols.AddIfNotNull(compilation.GetBestTypeByMetadataName("System.Xml.Serialization.XmlTypeAttribute"));
             XmlClassAttributeSymbols.AddIfNotNull(compilation.GetBestTypeByMetadataName("System.Xml.Serialization.XmlRootAttribute"));
+            XmlClassAttributeSymbols.AddIfNotNull(compilation.GetBestTypeByMetadataName("System.SerializableAttribute"));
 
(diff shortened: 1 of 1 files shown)


src/Meziantou.Analyzer/Rules/PreferReturningCollectionAbstractionInsteadOfImplementationAnalyzer.cs
```diff
@@ -49,6 +49,7 @@ public AnalyzerContext(Compilation compilation)
 
             XmlClassAttributeSymbols.AddIfNotNull(compilation.GetBestTypeByMetadataName("System.Xml.Serialization.XmlTypeAttribute"));
             XmlClassAttributeSymbols.AddIfNotNull(compilation.GetBestTypeByMetadataName("System.Xml.Serialization.XmlRootAttribute"));
+            XmlClassAttributeSymbols.AddIfNotNull(compilation.GetBestTypeByMetadataName("System.SerializableAttribute"));
```
meziantou, line 52: `SerializableAttribute` is for the Binary serializer, not for xml serializer. You can create a new property similar to `XmlClassAttributeSymbols`.

```
BinarySerializerClassAttributeSymbols.AddIfNotNull(compilation.GetBestTypeByMetadataName("System.SerializableAttribute"));
```

Micke3rd: its for binary and xml serialization
[Microsofts Doc](https://learn.microsoft.com/en-us/dotnet/api/system.serializableattribute?view=net-8.0) provides as [example](https://learn.microsoft.com/en-us/dotnet/api/system.serializableattribute?view=net-8.0#examples) the usage with SOAP

meziantou: If I'm not mistaken, while it's part of the documentation you mentioned, the XmlSerializer and SoapFormatter don't use this attribute. Only the binarySerializer requires it.

Micke3rd: oh, you're right ... so why do I use the attribute at all?  🙈😅

I hereby confirm that your code works fine as it is ^^

PR #743, opened 2024-07-16
New rule: MA0160 Use ContainsKey instead of TryGetValue when the second arg is a discard

### LINKED ISSUE #742: IDictionary.TryGetValue with a discard vs ContainsKey
**Is your feature request related to a problem? Please describe.**
From time to time, a developer will reuse/modify the code and will end up with code like this, mainly trying to use TryGetValue  instead of a ContainsKey check :
```cs
private ConcurrentDictionary<string, object> MyCollection = ...;
public static bool CheckExists(string libName)
{
    var exists = MyCollection.TryGetValue(libName, out _);
    return exists;
}
```

**Describe the solution you'd like**
the correct way should be to use a ContainsKey and not discard as it has some performance impact.

**Describe alternatives you've considered**
This is mainly caught during code reviews

**Additional context**
This should be check on different type implementing IDictionary/IReadOnlyDictionary

meziantou:
Fix #742

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -38,7 +38,7 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
     </PackageReference>
-    <PackageReference Include="Meziantou.Polyfill" Version="1.0.38">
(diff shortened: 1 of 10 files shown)


fforjan:
@meziantou Merci ! Thanks for implementing it !

PR #750, opened 2024-09-12
Update comparison doc to list CA1507 and S3376

drieseng:
Update comparison doc to list **CA1507** as equivalent of **MA0043**, and **S3376** as similar to **MA0058**.


drieseng:
@meziantou, can you confirm that *CA1507* is indeed an equivalent of **MA0043** and not just similar.

meziantou:
> can you confirm that CA1507 is indeed an equivalent of MA0043 and not just similar

There are similar but there are differences:
- MA0047 only report arguments of `ArgumentException` and derived types
- CA1507 reports for all arguments when the parameter name is `paramName` or `propertyName`. I haven't tested it, but `InvalidEnumArgumentException`'s parameter name is `argumentName`. Maybe it doesn't report it correctly.
- CA1507 suggest `nameof` for property name when the parameter name is `propertyName`

drieseng:
> > can you confirm that CA1507 is indeed an equivalent of MA0043 and not just similar
> 
> There are similar but there are differences:

I'll move **CA1507** / **MA0047** to the similar table.

> * MA0047 only report arguments of `ArgumentException` and derived types
> * CA1507 reports for all arguments when the parameter name is `paramName` or `propertyName`. I haven't tested it, but `InvalidEnumArgumentException`'s parameter name is `argumentName`. Maybe it doesn't report it correctly.

It indeed does not cover **InvalidEnumArgumentException**.

> * CA1507 suggest `nameof` for property name when the parameter name is `propertyName`

Not very useful, if you ask me.

I had disabled **MA0047** in favor of **CA1507**  in our code base, but I think I'll revert this.

PR #759, opened 2024-09-29
Added info for backport possibility for System.Threading.Lock

MarkCiliaVincenti:


meziantou:
Thanks for the contribution! However, I don't think I'll merge it:
- The rule only applies when `System.Threading.Lock` is available (.NET 9+ or already polyfilled)
- Other rules (IDExxxx, CAxxxx, MAxxxx) don't mentioned polyfills in similar cases. I prefer consistency.

PR #773, opened 2024-11-23
Implement Analyzer to add sealed keyword to public records

nodermatt:
The PublicRecordAnalyzer ensures that all public records in the codebase are annotated with the sealed keyword. This practice prevents other classes from inheriting from these records, which can help maintain the integrity and immutability of the data structures, reduce potential security risks, and improve performance by enabling certain compiler optimizations.

### DIFF
--- a/src/Meziantou.Analyzer.CodeFixers/Rules/PublicSealedRecordCodeFixer.cs
+++ b/src/Meziantou.Analyzer.CodeFixers/Rules/PublicSealedRecordCodeFixer.cs
@@ -1,0 +1,50 @@
+using System.Collections.Immutable;
+using System.Composition;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace Meziantou.Analyzer.Rules
+{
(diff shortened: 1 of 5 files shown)


meziantou:
There is already a rule to make classes public ([MA0053](https://github.com/meziantou/Meziantou.Analyzer/blob/main/docs/Rules/MA0053.md)). You need to set `MA0053.public_class_should_be_sealed = true` in the editorconfig to analyze public classes. Note that this may come with false-positive as Roslyn only analyze one project at a time. So, it doesn't have the full context of a solution. If a class is inherited in another assembly this cannot be detected.

nodermatt:
Hi @meziantou thanks for your quick response.
I did try that and found that rule MA0053 doesn't trigger for records even though they're classes as well. I added a unit test for the `ClassMustBeSealedAnalyzer` to demonstrate that.

What is your preferred course of action, extend MA0053 or go with a new one so the previous behaviour doesn't change?

meziantou:
MA0053 reports records with the following options:

````
[*.cs]
MA0053.public_class_should_be_sealed = true
MA0053.class_with_virtual_member_should_be_sealed = true
````

meziantou:
There is a fixer but it doesn't handle `record`. You need to update it:
- https://github.com/meziantou/Meziantou.Analyzer/blob/1b0e98c4d000bfbeeead7caa768f9e1c57361577/src/Meziantou.Analyzer.CodeFixers/Rules/ClassMustBeSealedFixer.cs#L29
- https://github.com/meziantou/Meziantou.Analyzer/blob/1b0e98c4d000bfbeeead7caa768f9e1c57361577/src/Meziantou.Analyzer.CodeFixers/Rules/ClassMustBeSealedFixer.cs#L45

PR #783, opened 2025-01-10
Align message to behavior for MA0157

### LINKED ISSUE #782: MA0157 phrasing is unclear
### Version of the Meziantou.Analyzer NuGet package

2.0.185

### Rule Identifier

MA0157

### Target Framework

net9.0

### C# Language version

C# 13

### Description

Updated the issue because at first I thought it was a bug, now I think I'm just confused :)

Based on the examples for MA0156
```csharp
// compliant
IAsyncEnumerable<string> FooAsync() => throw null;

// non-compliant
IAsyncEnumerable<string> Foo() => throw null;
```
and MA0157
```csharp
// compliant
IAsyncEnumerable<string> Foo() => throw null;

// non-compliant
IAsyncEnumerable<string> FooAsync() => throw null;
``` 
it looks like they're mutually exclusive. The first is for when you want an `Async` suffix for methods that return an `IAsyncEnumerable<T>` and the second is for when you don't want that. So we can choose which style we prefer by enabling either one of them.

If my understanding is correct, then the naming "MA0157 - Do not use 'Async' suffix when a method does not return IAsyncEnumerable<T>" seems confusing and doesn't represent what's going on. Because it's still about cases that _do_ return an `IAsyncEnumerable<T>`. Am I getting this right?

martindisch:
Closes #782.


meziantou:
Thanks for the improvement!

You'll need to fix the message in the test `Meziantou.Analyzer.Test.Rules.MethodsReturningAnAwaitableTypeMustHaveTheAsyncSuffixAnalyzerTests.IAsyncEnumerableWithSuffix`. Then, I'll merge it

PR #785, opened 2025-02-09
Fix MA0048 in case when first type is not first node

FrediKats:
Option "only_validate_first_type" now works more like "only_validate_first_node".

```
foreach (var node in root.DescendantNodesAndSelf(descendIntoChildren: node => !IsTypeDeclaration(node)))
```

This block of code returns CompilationUnit and `if (node.SpanStart < symbolNode.SpanStart)` always works as `if (0 < symbolNode.SpanStart)`.

PR contains:
- test to confirm this problem (namespace before type).
- fix with comparing only with other type declarations (skip nodes that is not type declaration)

### DIFF
--- a/tests/Meziantou.Analyzer.Test/Rules/FileNameMustMatchTypeNameAnalyzerTests.cs
+++ b/tests/Meziantou.Analyzer.Test/Rules/FileNameMustMatchTypeNameAnalyzerTests.cs
@@ -247,6 +247,19 @@ class [||]Test0<TKey, TValue>
               .ValidateAsync();
     }
 
+    [Fact]
+    public async Task MatchOnlyFirstType_TypeWithNamespaceDeclaration()
+    {
+        await CreateProjectBuilder()
+              .WithSourceCode(fileName: "Test0.cs", """
+                  namespace Sample;
+                  struct [||]Foo {}
+                  struct Bar {}
+                  """)
+              .AddAnalyzerConfiguration("MA0048.only_validate_first_type", "true")
+              .ValidateAsync();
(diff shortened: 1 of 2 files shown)


meziantou (approved):

tests/Meziantou.Analyzer.Test/Rules/FileNameMustMatchTypeNameAnalyzerTests.cs
```diff
@@ -247,6 +247,19 @@ struct Bar {}
               .ValidateAsync();
     }
 
+    [Fact]
+    public async Task MatchOnlyFirstType_TypeWithNamespaceDeclaration()
```
meziantou: You need to surround the test with:

```c#
#if CSHARP10_OR_GREATER
```

Locally, you use the latest version of Roslyn, but the CI runs against multiple versions of Roslyn, including 3.8 (needed for older versions of VS and .NET SDK) which does not support some new features.

FrediKats: Or I can you block scoped namespace. I guess it's better for quality control.
Thx for info, will update PR today.

meziantou: Or you can do both :) (there are never enough tests for a Roslyn analyzer)

meziantou (approved):

meziantou:
Thanks for your contribution 🚀

PR #826, opened 2025-08-25
Use Meziantou SDK

meziantou:


cnomen-visionplanner:
Hi @meziantou, could you please explain to me the reason for using your own SDK within this project, instead of directly referencing the Microsoft SDK as was done previously? Thanks in advance for your answer!

meziantou:
In short, it's an opiniated wrapper around the default SDK. It reduces duplication between my projects. The previous version (Meziantou.DotNet.CodingStandard) did very similar, but there were some limitations. So, I switch to an MSBuild SDK.

https://github.com/meziantou/Meziantou.NET.Sdk.

PR #869, opened 2025-09-13
docs: Add CA2259/MA0033 equivalent rule mapping

### LINKED ISSUE #845: MA0033 - Do not tag instance fields with ThreadStaticAttribute is handled by CA2259
- **Rule**: [MA0033](https://github.com/meziantou/Meziantou.Analyzer/blob/main/docs/Rules/MA0033.md) - Do not tag instance fields with ThreadStaticAttribute
- **Built-in Equivalent**: [CA2259](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2259) - Ensure ThreadStatic is only used with static fields

## Analysis

**MA0033** and **CA2259** serve the same purpose: ensuring that `ThreadStaticAttribute` is only applied to static fields, as applying it to instance fields is incorrect and can lead to unexpected behavior.

### MA0033 Coverage
- Detects ThreadStaticAttribute on instance fields
- Prevents incorrect usage of ThreadStatic

### CA2259 Coverage  
- Detects incorrect usage of ThreadStaticAttribute on non-static fields
- Official Microsoft rule for proper threading attribute usage
- Identical functionality to MA0033

## Recommendation

**CA2259** is the official Microsoft analyzer rule that provides **identical functionality** to MA0033. Since both rules target the same incorrect usage pattern, **MA0033 should be deprecated** in favor of CA2259.

## Migration Path

1. Enable **CA2259** in your analyzer configuration
2. Disable **MA0033** 
3. Update any rule suppressions from MA0033 to CA2259

## Related

- Parent issue: #835 - Consider deprecating rules that are now covered by built-in .NET analyzers

Meir017:
- Document that CA2259 (ThreadStatic only affects static fields) is equivalent to MA0033
- CA2259 was introduced in .NET 7 as the official Microsoft equivalent
- Both rules detect ThreadStaticAttribute applied to instance fields

Closes #845


Meir017:
closed in favor of #871

PR #884, opened 2025-10-05
New rule: MA0176 - Optimize guid creation

meziantou:


drieseng:
@meziantou, I'm not the only one keeping an eye on the commits in dotnet/runtime, it seems :-)

PR #928, opened 2025-11-13
MA0004: Auto-detect ASP.NET Core projects to skip ConfigureAwait requirement

steingran:
The UseConfigureAwaitAnalyzer (MA0004) now automatically detects ASP.NET Core projects by checking for referenced assemblies starting with "Microsoft.AspNetCore".

Previously, the analyzer only exempted specific ASP.NET Core framework types (Controllers, Razor Pages, Tag Helpers, etc.) from the ConfigureAwait requirement. Regular classes like services, repositories, and other business logic classes in ASP.NET Core projects were not detected and incorrectly required ConfigureAwait.

Since ASP.NET Core doesn't use a SynchronizationContext by default, ConfigureAwait is not needed for any classes in ASP.NET Core projects (except Blazor components, which maintain their own SynchronizationContext).

Changes:
- Added IsAspNetCoreProject property that checks compilation references
- Updated MustUseConfigureAwait logic to skip requirement for all classes in  ASP.NET Core projects (excluding Blazor components)
- Added 6 comprehensive tests covering regular classes, services, repositories, Blazor components, configuration overrides, and non-ASP.NET Core projects
- Existing configuration option (MA0004.report = always) still works to override the auto-detection when needed

All tests pass (6 new tests + existing test suite).

### DIFF
--- a/src/Meziantou.Analyzer/Rules/UseConfigureAwaitAnalyzer.cs
+++ b/src/Meziantou.Analyzer/Rules/UseConfigureAwaitAnalyzer.cs
@@ -59,6 +59,8 @@ public sealed class UseConfigureAwaitAnalyzer : DiagnosticAnalyzer
         private INamedTypeSymbol? AspNetCore_IFilterMetadata { get; } = compilation.GetBestTypeByMetadataName("Microsoft.AspNetCore.Mvc.Filters.IFilterMetadata");
         private INamedTypeSymbol? AspNetCore_IComponent { get; } = compilation.GetBestTypeByMetadataName("Microsoft.AspNetCore.Components.IComponent");
 
+        private bool IsAspNetCoreProject { get; } = compilation.ReferencedAssemblyNames.Any(a => a.Name.StartsWith("Microsoft.AspNetCore", StringComparison.Ordinal));
+
         public void AnalyzeAwaitOperation(OperationAnalysisContext context)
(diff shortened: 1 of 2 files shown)


meziantou:
> Updated MustUseConfigureAwait logic to skip requirement for all classes in ASP.NET Core projects (excluding Blazor components)

I'm not entirely convinced this is the right approach. Inside a Blazor component, you definitely don’t want the rule to trigger, but other classes used by the component could still benefit from it. Your change would disable the rule for other classes as they are inside a project having a dependency on ASP.NET Core.

My recommendation would be to disable the analyzer at the folder or project level using an `.editorconfig`. This avoids false positives and also improves build performance since the analyzer won't run on that project.

Another option would be to introduce a setting that automatically detects ASP.NET Core and disables the rule globally when the dependency is present. However, if this still requires adding a configuration entry in `.editorconfig`, then it’s essentially equivalent to just disabling the rule directly.

steingran:
Thank you for your feedback. Adding an .editorconfig file would definitely solve the problem with false positives regarding the MA0004 rule, but my intention here was to make it more "ready out-of-the-box" without having to configure the analyzer to give "correct" warnings/errors. But I do realize the solution I propose may be too broad, and disable the rule in scenarios where it would be beneficial. I will therefore close the PR

PR #1054, opened 2026-03-09
Add `roslyn5.0` packing

viceroypenguin:


.github/workflows/ci.yml
```diff
@@ -81,6 +81,9 @@ jobs:
     - run: dotnet build src/Meziantou.Analyzer/Meziantou.Analyzer.csproj --configuration Release /p:RoslynVersion=roslyn4.14 /p:Version=${{ needs.compute_package_version.outputs.package_version }} /bl:Meziantou.Analyzer.roslyn4.14.binlog
     - run: dotnet build src/Meziantou.Analyzer.CodeFixers/Meziantou.Analyzer.CodeFixers.csproj --configuration Release /p:RoslynVersion=roslyn4.14 /p:Version=${{ needs.compute_package_version.outputs.package_version }} /bl:Meziantou.Analyzer.CodeFixers.roslyn4.14.binlog
 
+    - run: dotnet build src/Meziantou.Analyzer/Meziantou.Analyzer.csproj --configuration Release /p:RoslynVersion=roslyn5.0 /p:Version=${{ needs.compute_package_version.outputs.package_version }} /bl:Meziantou.Analyzer.roslyn4.14.binlog
```
meziantou:
```suggestion
    - run: dotnet build src/Meziantou.Analyzer/Meziantou.Analyzer.csproj --configuration Release /p:RoslynVersion=roslyn5.0 /p:Version=${{ needs.compute_package_version.outputs.package_version }} /bl:Meziantou.Analyzer.roslyn5.0.binlog
```

meziantou (commented):
Can you update the directory.build.targets file to add an explicit case for roslyn5.0?

Frulfump:
There's no description for the PR, was this just missing but it still worked since v5 packages was still being pulled in or what's the purpose?

viceroypenguin:
@Frulfump The problem is that the roslyn5.0 binaries are not packaged into the nuget package, so roslyn v5 uses the 4.14 binaries currently. This PR updates the packaging to include roslyn5.0 binaries so v5 has access to certain v5 features (like the fix in #1053).

viceroypenguin:
@meziantou 
> Can you update the directory.build.targets file to add an explicit case for roslyn5.0?

I thought about that, but I didn't do it because then the default case doesn't make any sense. Do you have any thoughts on what you would prefer the default case to be? (Build fail, etc.)

meziantou:
> I thought about that, but I didn't do it because then the default case doesn't make any sense. Do you have any thoughts on what you would prefer the default case to be? (Build fail, etc.)

The default case targets the latest and can be updated at any time. This is mostly to allow testing with the latest version of Roslyn to detect bugs early. However, this is not a version that is shipped because it is mostly useless.

For consistency, I prefer having explicit case for the shipped versions (even if it is the same version of the default case at the moment)

viceroypenguin:
@meziantou Sounds good! Updated

viceroypenguin:
@meziantou sorry, watching the actions revealed one more missing case in the ci yml

PR #1061, opened 2026-03-24
MA0042 Generics fix

Griboedoff:
Hi
There are an issue with sync to async analyzer
It ignore cases when one of arguments is generic. 
Consider this:
```csharp
class Test
{
    public void A<T>(List<T> test)
    {
    }
    public async Task AAsync<T>(List<T> test, CancellationToken token = default)
    {
    }
}

class demo
{
    public async Task a()
    {
        new Test<int>().A([1]);
    }
}
```

When analyzer seraches for overload it compares `List<int> test` parameter from call with `List<T> test` from async overload. This comparison fails. So it is better to compare `OriginalDefinition` ot parameter type, it will return type without concrete type parameter if it is specified and do nothing otherwise

### DIFF
--- a/src/Meziantou.Analyzer/Internals/OverloadFinder.cs
+++ b/src/Meziantou.Analyzer/Internals/OverloadFinder.cs
@@ -123,7 +123,7 @@ internal sealed class OverloadFinder(Compilation compilation)
                 if (!options.IncludeObsoleteMembers && IsObsolete(method))
                     continue;
 
-                if (HasSimilarParameters(methodSymbol, method, options.AllowOptionalParameters, additionalParameterTypes))
+                if (OverloadFinder.HasSimilarParameters(methodSymbol, method, options.AllowOptionalParameters, additionalParameterTypes))
                     return method;
             }
         }
@@ -150,12 +150,18 @@ internal sealed class OverloadFinder(Compilation compilation)
         if (method.IsEqualTo(otherMethod))
             return false;
 
-        // The new method must have at least the same number of parameters as the old method, plus the number of additional parameters        
+        // The new method must have at least the same number of parameters as the old method, plus the number of additional parameters
         if (otherMethod.Parameters.Length - method.Parameters.Length < additionalParameterTypes.Length)
             return false;
 
         // If allowOptionalParameters is false, the new method must have exactly the same number of parameters as the old method
         if (!allowOptionalParameters && otherMethod.Parameters.Length - method.Parameters.Length != additionalParameterTypes.Length)
+            return false;
+
+        if (method.IsGenericMethod != otherMethod.IsGenericMethod)
+            return false;
+
+        if (method.IsGenericMethod && !HasSimilarGenericParametersInOrder(method, otherMethod))
             return false;
 
         // Most of the time, an overload has the same order for the parameters. Try to match them in order first (faster)
@@ -167,7 +173,8 @@ internal sealed class OverloadFinder(Compilation compilation)
                 var methodParameter = method.Parameters[i];
                 var otherMethodParameter = otherMethod.Parameters[j];
 
-                if (methodParameter.Type.IsEqualTo(otherMethodParameter.Type))
+                if (methodParameter.Type.IsEqualTo(otherMethodParameter.Type) ||
+                    ParametersWithGenericAreEqual(methodParameter, otherMethodParameter))
                 {
                     i++;
                     j++;
@@ -207,7 +214,8 @@ internal sealed class OverloadFinder(Compilation compilation)
                 var found = false;
                 for (var i = 0; i < otherMethodParameters.Length; i++)
                 {
-                    if (otherMethodParameters[i].Type.IsEqualTo(param.Type))
+                    if (otherMethodParameters[i].Type.IsEqualTo(param.Type) ||
+                        ParametersWithGenericAreEqual(param, otherMethodParameters[i]))
                     {
                         otherMethodParameters = otherMethodParameters.RemoveAt(i);
                         found = true;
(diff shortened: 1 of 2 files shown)


src/Meziantou.Analyzer/Internals/OverloadFinder.cs
```diff
@@ -207,7 +207,7 @@ public static bool HasSimilarParameters(IMethodSymbol method, IMethodSymbol othe
                 var found = false;
                 for (var i = 0; i < otherMethodParameters.Length; i++)
                 {
-                    if (otherMethodParameters[i].Type.IsEqualTo(param.Type))
+                    if (otherMethodParameters[i].Type.OriginalDefinition.IsEqualTo(param.Type.OriginalDefinition))
                     {
                         otherMethodParameters = otherMethodParameters.RemoveAt(i);
```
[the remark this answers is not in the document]
Griboedoff: this one looks like opposite of intended changes

src/Meziantou.Analyzer/Internals/OverloadFinder.cs
```diff
@@ -207,7 +207,7 @@ public static bool HasSimilarParameters(IMethodSymbol method, IMethodSymbol othe
                 var found = false;
                 for (var i = 0; i < otherMethodParameters.Length; i++)
                 {
-                    if (otherMethodParameters[i].Type.IsEqualTo(param.Type))
+                    if (otherMethodParameters[i].Type.OriginalDefinition.IsEqualTo(param.Type.OriginalDefinition))
```
meziantou: the new matching logic is broad enough that it should probably get a few negative tests in addition to the two positive tests already added.

I generated some cases locally, but maybe there are more cases needed:

```c#
[Fact]
public async Task GenericArgument_MultipleIncompatibleGenericArguments_ShouldNotReport()
{
    await CreateProjectBuilder()
          .WithSourceCode("""
            using System.Collections.Generic;
            using System.Threading;
            using System.Threading.Tasks;

            class Test
            {
                public void A(List<int> a, List<string> b)
                {
                }

                public Task AAsync<T>(List<T> a, List<T> b, CancellationToken token = default)
                {
                    return Task.CompletedTask;
                }
            }

            class Demo
            {
                public async Task M()
                {
                    new Test().A(new List<int>(), new List<string>());
                }
            }
            """)
          .ValidateAsync();
}
```

```c#
[Fact]
public async Task ExtensionMethod_GenericArgumentsIncompatible_ShouldNotReport()
{
    await CreateProjectBuilder()
          .WithSourceCode("""
            using System.Collections.Generic;
            using System.Threading;
            using System.Threading.Tasks;

            class Test
            {
            }

            static class TestExtensions
            {
                public static void A(this Test test, List<int> a, List<string> b)
                {
                }

                public static Task AAsync<T>(this Test test, List<T> a, List<T> b, CancellationToken token = default)
                {
                    return Task.CompletedTask;
                }
            }

            class Demo
            {
                public async Task M()
                {
                    new Test().A(new List<int>(), new List<string>());
                }
            }
            """)
          .ValidateAsync();
}
```

```c#
[Fact]
public async Task GenericArgument_DifferentGenericTypeDefinitions_ShouldNotReport()
{
    await CreateProjectBuilder()
          .WithSourceCode("""
            using System.Collections.Generic;
            using System.Threading;
            using System.Threading.Tasks;

            class Test
            {
                public void A(List<int> value)
                {
                }

                public Task AAsync<T>(IEnumerable<T> value, CancellationToken token = default)
                {
                    return Task.CompletedTask;
                }
            }

            class Demo
            {
                public async Task M()
                {
                    new Test().A(new List<int>());
                }
            }
            """)
          .ValidateAsync();
}
```

```c#
[Fact]
public async Task GenericArgument_NestedGenericIncompatibility_ShouldNotReport()
{
    await CreateProjectBuilder()
          .WithSourceCode("""
            using System.Collections.Generic;
            using System.Threading;
            using System.Threading.Tasks;

            class Test
            {
                public void A(List<List<int>> value)
                {
                }

                public Task AAsync<T>(List<List<T>> value, CancellationToken token = default)
                {
                    return Task.CompletedTask;
                }

                public Task AAsync(List<List<string>> value, CancellationToken token = default)
                {
                    return Task.CompletedTask;
                }
            }

            class Demo
            {
                public async Task M()
                {
                    new Test().A(new List<List<int>>());
                }
            }
            """)
          .ValidateAsync();
}
```

```c#
[Fact]
public async Task GenericArgument_SameOriginalDefinitionButDifferentTypeParameterMapping_ShouldNotReport()
{
    await CreateProjectBuilder()
          .WithSourceCode("""
            using System.Collections.Generic;
            using System.Threading;
            using System.Threading.Tasks;

            class Test
            {
                public void A<T1, T2>(Dictionary<T1, T2> value)
                {
                }

                public Task AAsync<T>(Dictionary<T, T> value, CancellationToken token = default)
                {
                    return Task.CompletedTask;
                }
            }

            class Demo
            {
                public async Task M()
                {
                    new Test().A<int, string>(new Dictionary<int, string>());
                }
            }
            """)
          .ValidateAsync();
}
```

meziantou:
Note: You can forget about the failing deploy step in the CI. This is expected for external contributors.

src/Meziantou.Analyzer/Internals/OverloadFinder.cs
Griboedoff: agreed, i'll check new tests

src/Meziantou.Analyzer/Internals/OverloadFinder.cs
Griboedoff: I've dug a bit deeper. I think this comparison always. 

[there is](https://github.com/dotnet/roslyn/blob/main/src/Compilers/CSharp/Portable/Symbols/Source/SourceParameterSymbolBase.cs#L42) `Equals` impl that was called when I debugged down. It checks `_containingSymbol` and they are always different because containing symbols are method symbols in this case.

You can check this yourself if you comment _slow search_ in line 200 and run this test
```csharp
[Fact]
public async Task SimpleArgument()
{
    await CreateProjectBuilder()
          .WithSourceCode("""
            using System;
            using System.Collections.Generic;
            using System.Threading;
            using System.Threading.Tasks;
            using System.Diagnostics;

            class Test
            {
                public void A(int i)
                {
                }

                public async Task AAsync(int i)
                {
                }
            }

            class demo
            {
                public async Task a()
                {
                    [|new Test().A(1)|];
                }
            }
            """)
          .ShouldFixCodeWith("""
            using System;
            using System.Collections.Generic;
            using System.Threading;
            using System.Threading.Tasks;
            using System.Diagnostics;

            class Test
            {
                public void A(int i)
                {
                }

                public async Task AAsync(int i)
                {
                }
            }

            class demo
            {
                public async Task a()
                {
                    await new Test().AAsync(1);
                }
            }
            """)
          .ValidateAsync();
}
```

Griboedoff: maybe there should be type comparison

`if (methodParameter.Type.IsEqualTo(otherMethodParameter.Type))`

meziantou:
Sorry, I'm pretty busy this week. I hope I will get more time next week to have a deeper look at this PR.

What would be useful to help is to have a list of tests of the dream state. I already provide a few tests, but I'm pretty sure there are more cases to handle. Then, we can check if we need some tradeoff to simplify the analyzer (or maybe let an agent figure out the code to match the spec).

Griboedoff:
Don't worry, i'll wait)

I've added tests for generic cases. To simplify comparison I decided that fixer can replace sync to async version only if _genericness_ of the methods are equal: same number of parameters, same constraints. It feels like a braking change a bit, but looks more intuitive to me

meziantou:
I took some time to think about more test cases. I extend the scope with numeric conversions (already in my radar) in addition to better support of generic types (thanks for reporting it!). The idea is to handle most cases, but not all of them. For instance, nested generics or compatible constraints based on the current type are excluded for simplicity.

To ease the discussion, I've created another draft PR. You can check the tests in https://github.com/meziantou/Meziantou.Analyzer/pull/1065/changes (do not look at the implementation, generated by copilot => not reviewed yet and surely need adjustments). If you agree with the test suite,
- Option 1: You can copy the tests in your PR and continue the implementation
- Option 2: I can have a look at it next week and proceed with my PR

Griboedoff:
I choose option 2:)

I checked tests in the PR, left few comments, but overall it covers all my needs. Thanks!

meziantou:
Lazy option ;)

Thanks for reviewing the other PR!

PR #1065, opened 2026-03-29
Improve MA0042 async overload matching coverage

meziantou:
## Why
`MA0042` overload detection was missing several realistic call-shape scenarios (generic inference, extension-vs-instance matching, ref-kind nuances, and numeric conversion compatibility), which caused false negatives in async context diagnostics.

## What changed
- Expanded overload matching in `OverloadFinder` to better support:
  - method-generic inference/constraint validation during comparison
  - `ref`/`out` strictness with `in`/by-value compatibility
  - extension/instance comparable parameter normalization
  - safer implicit numeric conversion compatibility (including `System.Half -> float/double`)
- Improved async-equivalent candidate discovery in `DoNotUseBlockingCallInAsyncContextAnalyzer` by broadening lookup around receiver/type context.
- Added a large async-context test matrix covering:
  - generic compatibility (positive and negative)
  - numeric widening/narrowing and floating-point precision-sensitive cases
  - instance vs extension method matching
  - argument forms (`default`, `null`, named args) and modifier edge cases
- Updated `DocumentationGenerator` to use `FindRequiredGitRepositoryRoot()` and bumped `Meziantou.Framework.FullPath` to `1.1.18`.

## Validation
- `dotnet test --filter "FullyQualifiedName~DoNotUseBlockingCallInAsyncContextAnalyzer_AsyncContextTests.Argument_ImplicitNumeric"` (passed)
- `dotnet test --filter "FullyQualifiedName~DoNotUseBlockingCallInAsyncContextAnalyzer_AsyncContextTests"` (passed)
- `dotnet run --project src/DocumentationGenerator` currently fails in this worktree with `Cannot find the current git folder` in this environment; code was updated toward required FullPath usage.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -48,6 +48,62 @@ The analyzer must use `IOperation` or `ISymbol` to analyze the content. Only fal
 Code snippets in tests must use raw string literals (`"""`) and must be minimized to only include the necessary code to reproduce the issue. Avoid including unnecessary code that does not contribute to the test case.
 When reporting a diagnostic, the snippet must use the `[|code|]` syntax or `{|id:code|}` syntax. Do not explicitly indicates lines or columns.
 
+### Code fixer best practice: validate before registering
+
+In `RegisterCodeFixesAsync`, validate **all** conditions that could prevent the fix from being applied **before** calling `context.RegisterCodeFix`. Do not register a code fix whose action would return the document unchanged.
+
+**Wrong** — registers the fix without validating whether it can be applied:
+```csharp
+public override async Task RegisterCodeFixesAsync(CodeFixContext context)
+{
+    var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+    var nodeToFix = root?.FindNode(context.Span, getInnermostNodeForTie: true);
+    if (nodeToFix is null)
+        return;
+
+    context.RegisterCodeFix(CodeAction.Create(title, ct => FixAsync(context.Document, nodeToFix, ct), equivalenceKey: title), context.Diagnostics);
+}
+
+private static async Task<Document> FixAsync(Document document, SyntaxNode nodeToFix, CancellationToken cancellationToken)
+{
+    if (nodeToFix is not BinaryExpressionSyntax binaryExpression)
+        return document; // Fix not applied — but it was already shown to the user!
+
+    var semanticModel = await document.GetSemanticModelAsync(cancellationToken).ConfigureAwait(false);
+    var mySymbol = semanticModel!.Compilation.GetBestTypeByMetadataName("System.SomeType");
+    if (mySymbol is null)
+        return document; // Fix not applied — but it was already shown to the user!
+    // ...
+}
+```
+
+**Correct** — validates all conditions first, then registers the fix:
+```csharp
+public override async Task RegisterCodeFixesAsync(CodeFixContext context)
+{
+    var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+    var nodeToFix = root?.FindNode(context.Span, getInnermostNodeForTie: true);
+    if (nodeToFix is not BinaryExpressionSyntax binaryExpression)
+        return;
+
+    var semanticModel = await context.Document.GetSemanticModelAsync(context.CancellationToken).ConfigureAwait(false);
+    if (semanticModel is null)
+        return;
+
+    if (semanticModel.Compilation.GetBestTypeByMetadataName("System.SomeType") is null)
+        return;
+
+    context.RegisterCodeFix(CodeAction.Create(title, ct => FixAsync(context.Document, binaryExpression, ct), equivalenceKey: title), context.Diagnostics);
+}
+
+private static async Task<Document> FixAsync(Document document, BinaryExpressionSyntax binaryExpression, CancellationToken cancellationToken)
+{
+    var editor = await DocumentEditor.CreateAsync(document, cancellationToken).ConfigureAwait(false);
+    // ... fix logic — all preconditions are guaranteed to hold
+    return editor.GetChangedDocument();
+}
+```
+
 ## Testing with different Roslyn versions
 
 This project supports multiple versions of Roslyn to ensure compatibility with different versions of Visual Studio and the .NET SDK. The supported Roslyn versions are configured in `Directory.Build.targets`:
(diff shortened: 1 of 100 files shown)


tests/Meziantou.Analyzer.Test/Rules/DoNotUseBlockingCallInAsyncContextAnalyzer_AsyncContextTests.cs
```diff
@@ -812,23 +808,964 @@ public async Task a()
...
+                {
+                    public void A(List<int> value) => throw null;
+                    public Task AAsync<T>(IEnumerable<T> value, CancellationToken token = default) => throw null;
+                }
+
+                class Demo
+                {
+                    public async Task M()
+                    {
+                        await new Test().AAsync(new List<int>());
```
meziantou: Should it report? List is assignable to IEnumerable, but maybe the sync method is more optimized for List.

tests/Meziantou.Analyzer.Test/Rules/DoNotUseBlockingCallInAsyncContextAnalyzer_AsyncContextTests.cs
```diff
@@ -812,23 +808,964 @@ public async Task a()
...
+                    public void A(List<List<int>> value) => throw null;
+                    public Task AAsync<T>(List<List<T>> value, CancellationToken token = default) => throw null;
+                    public Task AAsync(List<List<string>> value, CancellationToken token = default) => throw null;
+                }
+
+                class Demo
+                {
+                    public async Task M()
+                    {
+                        new Test().A(new List<List<int>>());
```
meziantou, line 933: Should it report?

tests/Meziantou.Analyzer.Test/Rules/DoNotUseBlockingCallInAsyncContextAnalyzer_AsyncContextTests.cs
```diff
@@ -812,23 +808,964 @@ public async Task a()
...
+                    public void A<T>(List<T> value) => throw null;
+                    public Task AAsync<T>(List<T> value, CancellationToken token = default)
+                        where T : class => throw null;
+                }
+
+                class Demo
+                {
+                    public async Task M()
+                    {
+                        new Test().A<int>(new List<int>());
```
meziantou, line 1127: Should it report if `A<string>`?

tests/Meziantou.Analyzer.Test/Rules/DoNotUseBlockingCallInAsyncContextAnalyzer_AsyncContextTests.cs
```diff
@@ -812,23 +808,964 @@ public async Task a()
...
+        await CreateProjectBuilder()
+              .WithSourceCode("""
+                using System.Collections.Generic;
+                using System.Threading;
+                using System.Threading.Tasks;
+
+                class Test
+                {
+                    public void A(List<int> value) => throw null;
+                    public Task AAsync<T>(IReadOnlyCollection<T> value, CancellationToken token = default) => throw null;
```
meziantou, line 1599: Add a test with IEnumerable
```
                    public void A(List<int> value) => throw null;
                    public Task AAsync<T>(IEnumerable<T> value, CancellationToken token = default) => throw null;
```

I don't think it should report in this case.

tests/Meziantou.Analyzer.Test/Rules/DoNotUseBlockingCallInAsyncContextAnalyzer_AsyncContextTests.cs
```diff
@@ -812,23 +808,1008 @@ public async Task a()
...
+    public async Task Argument_InModifierDifference_ShouldReport()
+    {
+        await CreateProjectBuilder()
+              .WithSourceCode("""
+                using System.Threading;
+                using System.Threading.Tasks;
+
+                class Test
+                {
+                    public void A(in int value) => throw null;
```
Griboedoff, line 1144: may be there should be separate analyzer without fix for `in`, `ref` and `out`, because modifiers are forbidden for async methods but you can rewrite it to async version

tests/Meziantou.Analyzer.Test/Rules/DoNotUseBlockingCallInAsyncContextAnalyzer_AsyncContextTests.cs
```diff
@@ -812,23 +808,1008 @@ public async Task a()
...
+                    {
+                        [|new Test().A((Half)1)|];
+                    }
+                }
+                """)
+              .ValidateAsync();
+    }
+
+    [Fact]
+    public async Task GenericArgument_NestedCompatibleGenericDefinitions_ShouldReport()
```
Griboedoff: `NestedCompatibleGenericDefinitions`

not nested, just compatible

tests/Meziantou.Analyzer.Test/Rules/DoNotUseBlockingCallInAsyncContextAnalyzer_AsyncContextTests.cs
```diff
@@ -812,23 +808,1008 @@ public async Task a()
...
+    {
+        await CreateProjectBuilder()
+              .WithSourceCode("""
+                using System.Collections.Generic;
+                using System.Threading;
+                using System.Threading.Tasks;
+
+                class Test
+                {
+                    public void A(Dictionary<int, string> value) => throw null;
```
Griboedoff, line 1624: arguments are incompatible without arity

tests/Meziantou.Analyzer.Test/Rules/DoNotUseBlockingCallInAsyncContextAnalyzer_AsyncContextTests.cs
```diff
@@ -812,23 +808,1008 @@ public async Task a()
...
+
+                class Value
+                {
+                    public static implicit operator Value(string value) => new Value();
+                }
+
+                class Test
+                {
+                    public void A(Value value) => throw null;
+                    public Task AAsync(Value value, CancellationToken token = default) => throw null;
```
Griboedoff, line 1751: parameters are compatible without implicit cast, may be sync method should be `A(string value)`

tests/Meziantou.Analyzer.Test/Rules/DoNotUseBlockingCallInAsyncContextAnalyzer_AsyncContextTests.cs
meziantou: The code should only accept the `in` modifier. `out` and `ref` would be incompatible. While it's an invalid modifier when you define an `async` method, an async method can use a method with a `in` parameter.

As the use-case is to replace method invocation, I think this is ok.
```c#
Task Foo(in int a) => throw null!;

async Task Bar(int a)
{
    await Foo(a); // Valid
}
```

Griboedoff: this sounds reasonable


But I'm talking about other case
```
bool TryA(int b, out Foo f) => throw null!;
Task(bool, Foo)> TryAAsync(int b) => throw null!;
```

I understand that it's a long shot for analyzer to guess this type of overload, but it can be usefull :)

meziantou: It would have more impact that just finding the _overload_. The code fix must be aware of it, not just using another method name, but assigning variables (it can be tricky depending on the context).

PR #1075, opened 2026-03-31
docs: require updating rule docs when rules change

meziantou:
## Why
When a rule implementation changes, its rule documentation can become stale if not updated in the same change. This adds an explicit instruction so analyzer behavior and `docs/Rules` stay aligned.

## What changed
- Updated `.github/copilot-instructions.md` to require updating the corresponding file under `docs/Rules/` whenever an existing rule is modified.

## Notes
- Scope is documentation guidance only (no analyzer/code behavior changes).


Frulfump:
Since AI isn't deterministic it could forget the instruction so this should probably use a hook instead.

PR #1114, opened 2026-05-03
Add global severity modes for MeziantouAnalysisMode

### LINKED ISSUE #1113: MSBuild property to set severity of all rules at once
**Is your feature request related to a problem? Please describe.**
According to what I understand of the README, you can only disable or enable the "default" set of the Analyzers by using the existing <MeziantouAnalysisMode> property with values "None" or "Default". As far as I understand there is no way to set all Analyzers to "Warning" for example using this way.

**Describe the solution you'd like**
Add more values to the <MeziantouAnalysisMode> property to set default severities of the ruleset.

**Describe alternatives you've considered**
Maybe there is an alternative way already other than setting them all one by one using the .editorconfig file that I am not aware of. Else just leave it as is.

meziantou:
This adds a simple way to set one baseline severity for all Meziantou rules from MSBuild. The existing `Default` and `None` modes were not enough when users want all rules to be warnings, errors, or suggestions.

## What changed
- Added new `MeziantouAnalysisMode` values:
  - `all-suggestions`
  - `all-warnings`
  - `all-errors`
- Wired these values in `build/Meziantou.Analyzer.props` to dedicated generated analyzer config files.
- Extended `DocumentationGenerator` to generate the new package config files:
  - `all-suggestions.editorconfig`
  - `all-warnings.editorconfig`
  - `all-errors.editorconfig`
- Updated README configuration docs and example to use the new mode names.

## Notes
- Existing `Default` and `None` behavior is unchanged.

Fixes: #1113


meziantou:
There is no way to set it in the editorconfig, but I don't think this is needed. Can you explain your use-case? Also, please open a new issue to explain your issue.

meziantou:
Roslyn analyzer configuration is driven by editorconfig files. You can change the default configuration by using standard values such as `dotnet_diagnostic.rule_id.severity`.
The MSBuild configuration allows you to provide a different baseline, such as disabling all rules by default. So, if you are only interested in a few rules, you can use the MSBuild property to set the baseline and then use the editorconfig to opt-in / opt-out individual rules. This is not possible using only editorconfig (at least I'm not aware of a solution managed by Roslyn).

meziantou:
- Having 2 locations to configure something doesn't seem problematic to me. 
- The MSBuild is completely optional. Like all Roslyn analyzers the editorconfig files is supported.

Also, you haven't answered the initial question. What's the actual issue you are trying to solve? Why is the MSBuild configuration problematic?

meziantou:
If you are aware of a solution supported by Roslyn to configure all rules from an analyzer, feel free to share it! I would be happy to remove the MSBuild configuration and replace it with a editorconfig-based solution.
But it must be supported by Roslyn to be performant. When you disable a rule using `dotnet_diagnostic.id.severity = none` (or similar using category), Roslyn doesn't run the code associated with the rule. To get this behavior, the solution must be supported by Roslyn, not a custom solution.

meziantou:
Changing the categories of existing rules would be a breaking change

PR #1213, opened 2026-07-04
MA0211: Flag redundant backing fields for primary constructor parameters

JonCanning:
### Summary

Adds a new analyzer (MA0211) that detects fields whose initializer merely copies a primary constructor parameter — e.g. `private readonly int _x = x;` in a class with a primary constructor `Foo(int x)`. The parameter is already captured as an implicit field, so the explicit backing field is redundant.

### What's included

**Analyzer** (`RedundantBackingFieldForPrimaryConstructorParameterAnalyzer`)
- Flags `FieldInitializer` operations where the value (after unwrapping implicit non-user-defined conversions) is a `ParameterReferenceOperation` to a primary constructor parameter
- Unwraps implicit widening/nullable-lifting conversions so `private readonly long _x = x;` (where `x` is `int`) is still flagged
- Suppresses when the field carries custom attributes (e.g. `[JsonPropertyName]`, `[DataMember]`) since attributes carry semantic meaning that doesn't transfer to the parameter (see [dotnet/roslyn#73899](https://github.com/dotnet/roslyn/issues/73899))
- Only active on C# 12+

**Code fixer** (`RedundantBackingFieldForPrimaryConstructorParameterFixer`) — two fixes:
- **Remove redundant field** — deletes the field declaration. Always offered. Batch-fixable.
- **Use parameter directly** — deletes the field and rewrites all references to the parameter name across the solution. Only offered when the field is `readonly` and the field type exactly matches the parameter type (no widening/nullable lifting), ensuring semantic equivalence.

**Shared helper** (`Internals/ConversionExtensions`)
- `UnwrapImplicitNonUserDefinedConversions` — extracts the implicit-conversion unwrapping logic used by both analyzer and fixer, eliminating duplication

**Tests** — 29 tests covering: basic detection, records, readonly structs, non-readonly fields, widening conversions, nullable lifting, user-defined conversions, attributed fields, both code fixes (including cross-document reference rewriting and `this.`-qualified references), and negative cases.

**Docs** — `docs/Rules/MA0211.md` with examples, rule configuration entries for all severity modes.

### Design decisions

- `Info` severity (suggestion) — the field is functional, just redundant
- `Design` category — matches MA0143 (primary constructor parameters should be readonly) and similar rules
- Attribute suppression avoids false positives where the field's attributes are load-bearing for serialization/interop
- The "Use parameter directly" fix is conservative: only offered when type-identical and readonly, to guarantee semantically-safe replacements

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,6 +229,7 @@ If you are already using other analyzers, you can check [which rules are duplica
 |[MA0208](https://github.com/meziantou/Meziantou.Analyzer/blob/main/docs/Rules/MA0208.md)|Usage|\[FixedAddressValueType\] fields must be value types|⚠️|✔️|❌|
 |[MA0209](https://github.com/meziantou/Meziantou.Analyzer/blob/main/docs/Rules/MA0209.md)|Performance|Use in keyword for in parameter|ℹ️|❌|✔️|
 |[MA0210](https://github.com/meziantou/Meziantou.Analyzer/blob/main/docs/Rules/MA0210.md)|Performance|Use in keyword to call the in overload|ℹ️|❌|✔️|
+|[MA0211](https://github.com/meziantou/Meziantou.Analyzer/blob/main/docs/Rules/MA0211.md)|Design|Do not create a backing field that merely copies a primary constructor parameter|ℹ️|✔️|✔️|
 
 <!-- rules -->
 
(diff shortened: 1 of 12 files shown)


meziantou:
I have a concern with the rule, mostly for readonly fields. Parameters cannot be readonly, whereas fields can. So, if you want the compiler to enforce readonly, using a readonly field is the only solution (no assignment, no ref, etc.).

Also, I would be curious to know how frequent the pattern is.
- I think most people migrate existing code base using the code fix for IDE0290, so there is no fields in this case as the code fix removes them
- If you write the code manually and you add the field explicitly, then there must be a reason for it.

I'm not fully sure yet but,
- I would make the rule disabled by default
- I think it should only report on non-readonly fields by default. Maybe add a configuration to support all fields.

JonCanning:
Excellent points, it's a rule I found useful for my own use case when using an agent to update legacy code, but maybe it doesn't stand up to general use.

PR #1223, opened 2026-07-21
Add MA0212: Use MemoryMarshal.GetReference instead of ref span[0]

meziantou:
## Summary

Adds rule **MA0212** that flags by-reference access to the first element (`[0]`) of `Span<T>`, `ReadOnlySpan<T>`, or arrays, and suggests safer `MemoryMarshal` APIs instead.

## Motivation

Using `ref span[0]` or `ref array[0]` to obtain a starting reference performs a bounds check and throws `IndexOutOfRangeException` when the buffer is empty. The correct pattern for performance-sensitive code is:

- `MemoryMarshal.GetReference(span)` — for `Span<T>` / `ReadOnlySpan<T>`
- `MemoryMarshal.GetArrayDataReference(array)` — for arrays (.NET 6+)

These APIs return a managed reference without bounds-checking, remaining safe even for empty buffers.

## What's detected

All by-reference forms are flagged:

| Form | Example |
|------|---------|
| `ref` local | `ref byte first = ref span[0];` |
| `ref` return | `return ref span[0];` |
| `ref` argument | `M(ref span[0]);` |
| `in` argument | `M(in span[0]);` |

Non-by-reference access (`span[0]` as a value) is **not** flagged.

## Code fix

The fixer rewrites the indexed expression to the appropriate `MemoryMarshal` call, preserving the surrounding `ref`/`in` keyword:

```csharp
// Before
M(ref span[0]);

// After
M(ref System.Runtime.InteropServices.MemoryMarshal.GetReference(span));
```

For arrays, the fix is only offered on .NET 6+ (where `GetArrayDataReference` is available); the diagnostic still fires on older targets.

## Tests

13 test cases: ref/in span and array arguments, ref-local and ref-return, ReadOnlySpan with in, constant-zero detection, no-diagnostic for non-zero/non-constant/value-access/custom-indexer, and array on .NET 5.

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,7 +230,7 @@ If you are already using other analyzers, you can check [which rules are duplica
 |[MA0209](https://github.com/meziantou/Meziantou.Analyzer/blob/main/docs/Rules/MA0209.md)|Performance|Use in keyword for in parameter|ℹ️|❌|✔️|
 |[MA0210](https://github.com/meziantou/Meziantou.Analyzer/blob/main/docs/Rules/MA0210.md)|Performance|Use in keyword to call the in overload|ℹ️|❌|✔️|
 |[MA0211](https://github.com/meziantou/Meziantou.Analyzer/blob/main/docs/Rules/MA0211.md)|Style|Use multi-line syntax for XML summary comments|ℹ️|❌|✔️|
-|[MA0212](https://github.com/meziantou/Meziantou.Analyzer/blob/main/docs/Rules/MA0212.md)|Usage|Use MemoryMarshal.GetReference instead of indexing at 0|⚠️|✔️|✔️|
+|[MA0212](https://github.com/meziantou/Meziantou.Analyzer/blob/main/docs/Rules/MA0212.md)|Usage|Use MemoryMarshal.GetReference instead of indexing at 0|⚠️|❌|✔️|
 
 <!-- rules -->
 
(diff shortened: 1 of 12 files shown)


xtqqczze:
> These APIs return a managed reference without bounds-checking, remaining safe even for empty buffers.

I don't think it is accurate to describe `MemoryMarshal` APIs as "safe". In the case of `GetReference`, it is illegal to dereference the returned reference when the span is empty.

See also: https://github.com/dotnet/runtime/issues/127098

cc: @EgorBo

xtqqczze:
It would be more useful to have an analyzer to use `ref span[0]` instead of `MemoryMarshal.GetReference`.

meziantou:
The rule is disabled by default and not intended to be enable by default anytime in the future. This is helpful to analyze a full code base and find location to review manually when needed.

If you are using MemoryMarshal, Unsafe or similar class, you are not in the _safe_ area.

Maybe the description should be updated (mostly AI generated) to be more explicit.

EgorBo:
> If you are using MemoryMarshal, Unsafe or similar class, you are not in the safe area.

There are quite a few perfectly safe APIs in those (even in Unsafe), but yeah, `MemoryMarshal.GetReference` is among the most unsafe ones.

From my experience, JIT typically has hints in surrounded code (esp. after inlining) that an array is not empty, so the bound check from `ref span[0]` would be eliminated.

EgorBo:
@meziantou I've looked into the rest of the analyzers and they look very reasonable, especially `UseInlineArrayInsteadOfFixedBuffer` (which actually does the opposite of what this one does: `UseInlineArrayInsteadOfFixedBuffer` introduces bounds checks for code that didn't have them for better safety). Also, in .NET 12.0 `MemoryMarshal.GetReference` will start requiring an `unsafe {}` context (not sure if by default or via opt-in rules, but still), just FYI. In general, a deterministic NRE is perfectly safe vs potentially UB managed reference produced by `GetReference`.

PR #1291, opened 2026-08-17
Add MA0216: Remove unnecessary closed modifier

meziantou:
## What

New rule **MA0216 — Remove unnecessary closed modifier**.

C# 15 introduces closed hierarchies. A `closed` type is implicitly abstract, so a `closed` class or record with no derived type can never be instantiated — it cannot be used at all.

```csharp
closed class Sample { }             // MA0216

closed class Sample { }             // ok
class Derived : Sample { }
```

Category `Design`, default severity `Info`.

## No code fix

Deliberately none. Every way to resolve the diagnostic changes the meaning of the code:

- removing the modifier makes the type instantiable **and** inheritable
- replacing it with `sealed` makes it instantiable
- replacing it with `abstract` lets any type derive from it
- adding a derived type completes the hierarchy

Since the type has no derived type, it also has no existing usage a fix would need to preserve. The choice belongs to the developer.

## How

- [`RemoveUnnecessaryClosedModifierAnalyzer`](https://github.com/meziantou/Meziantou.Analyzer/blob/feature/epic-yalow-99469e/src/Meziantou.Analyzer/Rules/RemoveUnnecessaryClosedModifierAnalyzer.cs): symbol action on `NamedType` using `ITypeSymbol.IsClosed` and `GetClosedDerivedTypeInfo()`. No diagnostic when `IsComplete` is `false`.
- The diagnostic is reported on every declaration carrying the token, since `closed` can be set on a single part of a partial type.

## Notes for reviewers

- `ITypeSymbol.IsClosed`, `GetClosedDerivedTypeInfo`, and `SyntaxKind.ClosedKeyword` exist only in Roslyn 5.9 and are marked experimental (`RSEXPERIMENTAL006`). The analyzer is wrapped in `#if ROSLYN_5_9_OR_GREATER` with scoped pragma suppressions, following the existing `UseStringComparerAnalyzer` pattern. The rule compiles out on roslyn4.8/5.0/5.6.
- Tests require `LanguageVersion.Preview` (there is no `LanguageVersion.CSharp15` yet) and the net11.0 reference pack, which provides the `System.Runtime.CompilerServices.IsClosedTypeAttribute` the compiler needs to emit `closed` types.

## Validation

- 12 tests covering records, nested types, generics, partial declarations, a `closed` type derived from another `closed` type, a type with an abstract member, and derived types in another file.
- Full suite: 3620/3620 on the default Roslyn, 3511/3511 on roslyn4.8 (delta is the guarded tests). Builds clean on roslyn4.8/5.0/5.6.
- `dotnet run --project src/DocumentationGenerator` run until stable.


docs/Rules/MA0216.md
```diff
@@ -0,0 +1,30 @@
...
+<!-- sources -->
+Source: [RemoveUnnecessaryClosedModifierAnalyzer.cs](https://github.com/meziantou/Meziantou.Analyzer/blob/main/src/Meziantou.Analyzer/Rules/RemoveUnnecessaryClosedModifierAnalyzer.cs)
+<!-- sources -->
+
+A `closed` type is implicitly abstract, so it can only be instantiated through one of its derived types. When no type derives from it, the type cannot be used at all.
+
+The rule is only reported when the compiler can determine the complete list of derived types.
+
+There is no code fix, as every way to resolve the diagnostic changes the meaning of the code: removing the modifier makes the type instantiable and inheritable, replacing it with `sealed` makes it instantiable, replacing it with `abstract` allows any type to derive from it, and adding a derived type completes the hierarchy.
+
```
meziantou:
```suggestion
```

PR #1302, opened 2026-08-19
Use the Meziantou.Framework.Roslyn package

meziantou:
The repository maintained its own copies of the Roslyn compilation constants and of a large set of analyzer helpers. Both are now provided by [`Meziantou.Framework.Roslyn`](https://github.com/meziantou/Meziantou.Framework/blob/main/src/Meziantou.Framework.Roslyn/readme.md) 1.0.1, which compiles the helper sources into the consuming project. Net effect: **-2199 lines**.

## Constants

The `DefineConstants` blocks are gone from the root `Directory.Build.targets`; the package derives `ROSLYN_*_OR_GREATER` and `CSHARP*_OR_GREATER` from the version of the referenced `Microsoft.CodeAnalysis.*` packages. Only the version-specific `NoWarn` entries and the unsupported-version guard remain. Verified with a temporary probe file that `roslyn5.9` gets `ROSLYN_5_9_OR_GREATER`/`CSHARP15_OR_GREATER` and `roslyn4.8` gets neither.

The test project references the package with `ExcludeAssets="contentfiles"` — it only needs the constants, since the helpers are already visible through `InternalsVisibleTo`. Compiling them there too would make every reference ambiguous.

Two constant mappings changed, and both are worth a look:

- **`ROSLYN5_0`** (bare, no `_OR_GREATER`) no longer exists. Its two usages in `ProjectBuilder` become `ROSLYN_5_0_OR_GREATER`, which is what they meant when written in #916 — 5.0 was the newest version then, and the `Microsoft.CodeAnalysis.NetAnalyzers` selection had since silently stopped applying to 5.6 and 5.9.
- **`CSHARP15_OR_GREATER`** starts at Roslyn 5.6 in the package, but started at 5.9 here. `closed` types and union types are only supported from 5.9 and their analyzer code is already guarded with `ROSLYN_5_9_OR_GREATER`, so the corresponding tests now use the same guard. The `UseStringComparer` collection-expression code genuinely works on 5.6 and keeps `CSHARP15_OR_GREATER`, gaining coverage there.

## Helpers

Deleted outright: `CompilationExtensions`, `LanguageVersionExtensions`, `LocalDataFlowAnalysis`, `LocationExtensions`, `NamespaceSymbolExtensions`, `SuppressorHelpers`, and the whole `DiagnosticReporter` set (`ContextExtensions`, the five `Diagnostic*ReportOptions`, and the T4 template that generated them).

`MethodSymbolExtensions`, `OperationExtensions`, `SymbolExtensions` and `TypeSymbolExtensions` keep only what the package does not provide — `GetChildOperations`, `HasArgumentOfType`, `LookupAvailableSymbols`, `IsEqualTo`, `IsConst`, the top-level-statement helpers, `IsUnitTestMethod`/`IsUnitTestClass`, and the union-type helpers. Call sites use the package names (`UnwrapImplicitConversions`, `IsOrInheritsFrom`, `GetFirstAttribute`, `IsCSharp*OrGreater`, `FindNode`, …).

### Behaviour change worth reviewing

The package's `GetSymbolType` maps `INamedTypeSymbol`/`ITypeParameterSymbol` to themselves where the local one returned `null`. `UseAnOverloadThatHasCancellationToken` and `UseAnOverloadThatHasTimeProvider` feed it the symbols available at a location, so they started walking the members of every type name in scope (a test caught this: extra `HttpContext.RequestAborted`, `Test.MyCancellationToken`, … in the message). Both loops now skip `ITypeSymbol` alongside the existing `IMethodSymbol` guard — only values can hold a token or a provider. `DebuggerDisplayAttributeShouldContainValidExpressions` already filtered `INamespaceOrTypeSymbol` before the call, so it is unaffected.

## Verification

- All five Roslyn versions build with 0 warnings and 0 errors.
- All tests pass: 3528 / 3575 / 3601 / 3625 / 3658 for 4.8 / 4.14 / 5.0 / 5.6 / 5.9. The 5.6 total drops by 21 because the `closed`/union tests are now correctly excluded there.
- `dotnet run --project src/DocumentationGenerator` exits 0 with no documentation changes.
- `dotnet pack` produces all five analyzer/code-fixer pairs, and the nuspec has no leaked dependency (`PrivateAssets="all"`).

`AGENTS.md` is updated to say where the constants now come from and to note the `CSHARP15_OR_GREATER` caveat.


Directory.Build.targets
```diff
@@ -17,48 +17,40 @@
...
     </When>
   </Choose>
 
-  <!-- Without this, a project named after an unknown Roslyn version would silently compile without any ROSLYN_* constant -->
+  <!-- The ROSLYN_* and CSHARP*_OR_GREATER constants come from the Meziantou.Framework.Roslyn package, which derives them
+       from the version of the Microsoft.CodeAnalysis.* packages. Only the version specific warnings remain here.
+       Without this check, a project named after an unknown Roslyn version would silently compile with the wrong warnings -->
   <Target Name="EnsureRoslynVersionIsSupported" BeforeTargets="BeforeBuild" Condition="'$(RoslynVersionSupported)' != 'true'">
     <Error Text="The Roslyn version '$(RoslynVersion)' is not supported. Add it to the Directory.Build.targets of the repository root." />
   </Target>
```
meziantou: I don't think this is needed. There is no way to use an unsupported version now.

Directory.Build.targets
```diff
@@ -17,48 +17,40 @@
...
         <RoslynVersionSupported>true</RoslynVersionSupported>
-        <DefineConstants>$(DefineConstants);ROSLYN5_6;ROSLYN_4_2_OR_GREATER;ROSLYN_4_4_OR_GREATER;ROSLYN_4_5_OR_GREATER;ROSLYN_4_6_OR_GREATER;ROSLYN_4_8_OR_GREATER;ROSLYN_4_10_OR_GREATER;ROSLYN_4_14_OR_GREATER;ROSLYN_5_0_OR_GREATER;ROSLYN_5_6_OR_GREATER</DefineConstants>
-        <DefineConstants>$(DefineConstants);CSHARP9_OR_GREATER;CSHARP10_OR_GREATER;CSHARP11_OR_GREATER;CSHARP12_OR_GREATER;CSHARP13_OR_GREATER;CSHARP14_OR_GREATER</DefineConstants>
         <NoWarn>$(NoWarn);CS0618</NoWarn>
       </PropertyGroup>
     </When>
 
     <When Condition="$(RoslynVersion) == 'roslyn5.9'">
       <PropertyGroup>
         <RoslynVersionSupported>true</RoslynVersionSupported>
```
meziantou: With the other comment, this property has no usage

Directory.Build.targets
```diff
@@ -17,48 +17,40 @@
...
         <NoWarn>$(NoWarn);nullable</NoWarn>
       </PropertyGroup>
     </When>
 
     <When Condition="$(RoslynVersion) == 'roslyn4.14'">
       <PropertyGroup>
         <RoslynVersionSupported>true</RoslynVersionSupported>
-        <DefineConstants>$(DefineConstants);ROSLYN_4_8;ROSLYN_4_2_OR_GREATER;ROSLYN_4_4_OR_GREATER;ROSLYN_4_6_OR_GREATER;ROSLYN_4_8_OR_GREATER;ROSLYN_4_14_OR_GREATER</DefineConstants>
-        <DefineConstants>$(DefineConstants);CSHARP9_OR_GREATER;CSHARP10_OR_GREATER;CSHARP11_OR_GREATER;CSHARP12_OR_GREATER;CSHARP13_OR_GREATER</DefineConstants>
         <NoWarn>$(NoWarn);nullable</NoWarn>
```
meziantou: Is there a better way to keep nullable only for the latest version?

Directory.Build.targets
meziantou: Removed, along with the `RoslynVersionSupported` property.

It existed because the `ROSLYN_*` constants were defined in this file: a project named after a version with no `When` block compiled without any of them, silently. Now that the constants come from the package and `RoslynPackageVersion` is derived from the project name, an unknown version fails the restore instead — no `Microsoft.CodeAnalysis.*` package matches.

Directory.Build.targets
meziantou: Right, removed. The whole `Choose` block went with it: what was left was two `NoWarn` entries, now plain conditional `PropertyGroup`s.

Directory.Build.targets
meziantou: Inverted the condition so the versions are not listed at all:

```xml
<PropertyGroup Condition="'$(RoslynVersion)' != '$(DefaultRoslynVersion)'">
  <NoWarn>$(NoWarn);nullable</NoWarn>
</PropertyGroup>
```

`DefaultRoslynVersion` is a new property of the root `Directory.Build.props` that also feeds the default value of `RoslynVersion`, so there is a single place to bump. Adding a Roslyn version no longer requires deciding whether it needs the `nullable` opt-out.

One behaviour change worth confirming: `roslyn5.0` and `roslyn5.6` reported nullable warnings before and no longer do — only `roslyn5.9` does. That is what "only for the latest version" means, and it matches the reason for the suppression (the Roslyn nullable annotations change between versions and the code is written against the default one), but say the word if you would rather keep them reported for 5.0 and 5.6.

Resolved `NoWarn` per version is now:

| version | NoWarn |
|---|---|
| roslyn4.8 | `CA1014;CS1573;CS1591;nullable` |
| roslyn4.14 | `CA1014;CS1573;CS1591;nullable` |
| roslyn5.0 | `CA1014;CS1573;CS1591;nullable` |
| roslyn5.6 | `CA1014;CS1573;CS1591;nullable;CS0618` |
| roslyn5.9 | `CA1014;CS1573;CS1591` |

Directory.Build.targets
```diff
@@ -1,67 +1,27 @@
...
+  <PropertyGroup Condition="'$(RoslynVersion)' != '$(DefaultRoslynVersion)'">
+    <NoWarn>$(NoWarn);nullable</NoWarn>
+  </PropertyGroup>
 
-  <!-- Without this, a project named after an unknown Roslyn version would silently compile without any ROSLYN_* constant -->
-  <Target Name="EnsureRoslynVersionIsSupported" BeforeTargets="BeforeBuild" Condition="'$(RoslynVersionSupported)' != 'true'">
-    <Error Text="The Roslyn version '$(RoslynVersion)' is not supported. Add it to the Directory.Build.targets of the repository root." />
-  </Target>
+  <PropertyGroup Condition="'$(RoslynVersion)' == 'roslyn5.6'">
+    <NoWarn>$(NoWarn);CS0618</NoWarn>
```
meziantou: I think you can remove this suppression

meziantou: Removed. Building every Roslyn version in both Debug and Release reports no CS0618 warning, so it was dead configuration hiding a genuine obsolete-API usage — the same finding as #1290, which removed the entries for the other versions but left this one behind.

`Directory.Build.targets` now holds nothing specific to a single Roslyn version: the only remaining conditions are `RoslynVersion != DefaultRoslynVersion` and `TargetFramework == netstandard2.0`.

Re-verified after the change: all five versions build with 0 warnings in Debug and Release, and the tests still pass (3528 / 3575 / 3601 / 3625 / 3658).

tests/Meziantou.Analyzer.Test/Directory.Build.props
```diff
@@ -11,6 +11,8 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <!-- The helpers are [Embedded], so the ones of Meziantou.Analyzer are not visible here even through InternalsVisibleTo: compile them -->
```
meziantou:
```suggestion
```

PR #1470, opened 2026-09-11
Do not report MA0148/MA0149 when the constant does not convert to the operand type

meziantou:
## Problem

MA0148/MA0149 rewrite `value == 1L` (with `int value`) into `value is 1L`, which fails with **CS0266**. The binary equality operator promotes the `int` operand to `long`, but the fixer copies the operand syntax into an `is` expression, dropping the implicit conversion. Constant patterns are matched against the operand's own type, so the `long` constant is not valid there.

The existing guard only excluded implicit user-defined conversions, not built-in numeric promotion.

## Fix

When the comparison implicitly converts the operand, the analyzer and the fixer now check that the constant is implicitly convertible (without user-defined conversion) to the type of the operand as written in source (or its underlying type when nullable). If not, the comparison is not reported nor rewritten. This check replaces `HasImplicitUserDefinedConversion`, which it covers.

- Still reported/fixed: `byteValue == 1` → `byteValue is 1` (same for `short`, `ushort`, `byte?`)
- No longer reported: `int` vs `1L` / `1.0` / `1m`, `int?` vs `1L`, `float` vs `0.1`, `byte` vs `300`, `char` vs `65`
- Merged chains keep the valid terms only: `value == 1 || value == 2L` → `value is 1 || value == 2L`

The fixer shares the check for both single comparisons and merged `||`/`&&` candidates, so a fix triggered on a valid term can no longer merge an invalid one.

### Note for reviewers

The operand type is taken from `SemanticModel.GetTypeInfo(operandSyntax).Type` rather than by unwrapping the conversion operations: Roslyn 4.8 represents `(DayOfWeek?)1` as a chain of implicit conversions down to the `int` literal, so unwrapping overshoots and broke the existing `EqualityComparison_NullableEnum` test on that version.

## Tests

- New tests in `UsePatternMatchingForEqualityComparisonsAnalyzerTests` for non-convertible constants, convertible promoted operands, and mixed/merged comparisons (verified that 9 of the new cases fail without the fix).
- MA0148/MA0149 tests pass on Roslyn 4.8, 4.14, 5.0, 5.6 and 5.9 (46/46 each).
- Full test suite passes on Roslyn 5.9 (4394/4394).
- `docs/Rules/MA0148.md` and `docs/Rules/MA0149.md` document the new exclusion; `DocumentationGenerator` reports no further changes.

### DIFF
--- a/src/Meziantou.Analyzer/Rules/UsePatternMatchingForEqualityComparisonsCommon.cs
+++ b/src/Meziantou.Analyzer/Rules/UsePatternMatchingForEqualityComparisonsCommon.cs
@@ -1,4 +1,8 @@
-﻿namespace Meziantou.Analyzer.Rules;
+﻿using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace Meziantou.Analyzer.Rules;
+
 internal static class UsePatternMatchingForEqualityComparisonsCommon
 {
     public static bool IsNull(IOperation operation)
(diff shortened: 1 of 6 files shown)


src/Meziantou.Analyzer/Rules/UsePatternMatchingForEqualityComparisonsCommon.cs
```diff
@@ -20,16 +24,23 @@ public static bool IsConstantLiteral(IOperation operation)
...
+            return true;
 
-            operation = conversionOperation.Operand;
-        }
+        var semanticModel = expressionOperation.SemanticModel;
+        if (semanticModel is null || expressionOperation.Syntax is not ExpressionSyntax expression || constantOperation.Syntax is not ExpressionSyntax constantExpression)
+            return false;
 
-        return false;
+        var operandType = semanticModel.GetTypeInfo(expression, cancellationToken).Type;
```
meziantou: Do we need to use the semantic model? Can't we use the operation.Type directly?

meziantou: Good catch for the operand type: it is now read from the operation tree (`conversionOperation.Operand.Type`), and `GetTypeInfo` is gone. Tests pass on Roslyn 4.8, 4.14, 5.0, 5.6 and 5.9.

The semantic model is still needed for `ClassifyConversion`, though: the conversion of the constant depends on its value, not only on its type. `byteValue == 1` must stay reported (`byteValue is 1` compiles thanks to the implicit constant expression conversion), while `byteValue == 300` must not. `Compilation.ClassifyConversion(int, byte)` returns an explicit conversion for both, so only the expression-based overload of the semantic model distinguishes them.

